PR #13, opened 2015-05-23
Changed Android toolchain to work with hunter

pretyman:
Hi Ruslan,

I have made small changes to get the Android toolchain working with hunter.

To whoever wants to try it out: please be aware that the ANDROID_NDK, as other variables must be set as environment variables, and *not* be passed to cmake as -DANDROID_NDK=..., because when hunter calculated the toolchain's sha1 (at modules/hunter_calculate_toolchain_sha1.cmake) it does not forward these variables to the cmake child process, making the sha1 calculation fail and the whole cmake process abort.

What I do is that I put the variables in front of the commanline calling cmake, for example:
ANDROID_NDK=${ANDROID_NDK} ANDROID_ABI="armeabi-v7a with NEON" ANDROID_STL=gnustl_static ANDROID_NATIVE_API_LEVEL=13 cmake -DCMAKE_TOOLCHAIN_FILE=${POLY_HOME}/os/android.toolchain.cmake

Regards


ruslo:
Thanks. I will take a look closer later, probably Hunter need to be upgraded for forwarding ANDROID_* variables to resolve environment/CMake variables issue

PR #14, opened 2015-05-26
Android: updated the toolchain

pretyman:
Updated the toolchain to set CMAKE_ANDROID_API.
CMAKE_{C,CXX}_FLAGS now contain the include directories of STL
Replaced tabs with 2 spaces.


pretyman:
yes, using `ANDROID` makes it more clear and renders the `string(COMPARE...)` redundant, hence not needed.

PR #16, opened 2015-05-27
Update the Android toolchain

pretyman:
Corrected the CMAKE_ANDROID_API (I did a wrong commit previously, sorry
about that)

Set CMAKE_COMPILER_IS_GNUCXX which is used by hunter's Boost
COMPONENT's scheme (however, CMake 3.2 describes this as the "old"
compiler id variable)

Set CMAKE_CXX_COMPILER_ID, which I belive is the "new" compiler id

Set CMAKE_CXX_COMPILER_ID_RUN, which I don't know if it is correct,
this used to be set by the CMAKE_FORCE_CXX_COMPILER call that was
removed.

### DIFF
--- a/os/android.toolchain.cmake
+++ b/os/android.toolchain.cmake
@@ -1169,6 +1169,10 @@ endif()
 # Force set compilers because standard identification works badly for us
 #include( CMakeForceCompiler ) AWP: I've disabled this to work with hunter
 #CMAKE_FORCE_C_COMPILER( "${CMAKE_C_COMPILER}" GNU ) AWP: I've disabled this to work with hunter
+set(CMAKE_C_COMPILER_ID_RUN TRUE)
+set(CMAKE_C_COMPILER_ID GNU)
+set(CMAKE_COMPILER_IS_GNUCC 1) # AWP: Set old compiler ID variable, as per CMake 3.2 Modules/CMakeForceCompiler.cmake
+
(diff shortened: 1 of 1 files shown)


ruslo:
What do you think about sending this changes [upstream](https://github.com/taka-no-me/android-cmake)? In case if we want to make an updates in future it would be much easier just to copy/paste the code than doing manual merge

pretyman:
Yes, it is a good idea.
I will get the upstream and compare. I'll check what is missing and ask their input on how can it better achieved.

ruslo:
> Set CMAKE_COMPILER_IS_GNUCXX

I've tested it with current toolchain and seems that this variables are set:

```
CMAKE_C_COMPILER_ID_RUN: 1
CMAKE_C_COMPILER_ID: GNU
CMAKE_COMPILER_IS_GNUCC: 1
CMAKE_CXX_COMPILER_ID_RUN: 1
CMAKE_CXX_COMPILER_ID: GNU
CMAKE_COMPILER_IS_GNUCXX: 1
```

Also newer version of `android.toolchain.cmake` already has `CMAKE_ANDROID_API` as you already mentioned

ruslo:
I guess we don't need this anymore?

pretyman:
Nope, polly has the updated android.toolchain.cmake. You can close this.

Thanks

PR #21, opened 2015-06-17
Added more android toolchains

headupinclouds:
Existing polly toolchains have not specified ANDROID_TOOLCHAIN_NAME in the filename, which implies the default of arm-linux-androideabi-4.9 in the android.toolchain.cmake file.  These will always be tied to the default, which is probably convenient for many scenarios.

I've added a few new toolchains here that explicitly specify the compiler/toolchain.  In some cases this is identical to the current default, but it will guarantee the same build type for any future releases of android.toolchain.cmake.

I suppose some versions may also want to specify ANDROID_STL as libc++ becomes more widely supported.


ruslo:
@headupinclouds Thanks

PR #23, opened 2015-06-27
Bugfix/make-framework-remove-cmake-directory

headupinclouds:
This simple change just prevents errors of the form:

"ValueError: list.remove(x): x not in list"

when the installation steps doesn't create a lib/cmake directory.


ruslo:
Thanks you.

PR #28, opened 2015-07-06
new toolchains + fixes 

headupinclouds:
* added 8.2 and 8.4 iOS SDK toolchains w/ libc++
* fix typo in ANDROID_TOOLCHAIN_NAME for arm-linux-androidabi-clang3.5 toolchain
* change dSYM to FMWK in dynamic framework info.plist template (see Core Foundation Keys CFBundlePackageType in https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html)

### DIFF
--- a/ios-8-2-libcxx.cmake
+++ b/ios-8-2-libcxx.cmake
@@ -1,0 +1,38 @@
+# Copyright (c) 2015, Ruslan Baratov
+# All rights reserved.
+
+if(DEFINED POLLY_IOS_8_2_LIBCXX_CMAKE_)
+  return()
+else()
+  set(POLLY_IOS_8_2_LIBCXX_CMAKE_ 1)
+endif()
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_clear_environment_variables.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+set(IOS_SDK_VERSION 8.2)
+set(POLLY_XCODE_COMPILER "clang")
+polly_init(
+    "iOS ${IOS_SDK_VERSION} Universal (iphoneos + iphonesimulator) / \
+${POLLY_XCODE_COMPILER} / LLVM Standard C++ Library (libc++)  \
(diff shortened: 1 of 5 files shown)


ios-8-2-libcxx.cmake
```diff
@@ -0,0 +1,38 @@
...
+set(MACOSX_BUNDLE_GUI_IDENTIFIER com.example)
+set(CMAKE_MACOSX_BUNDLE YES)
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
+
+set(IPHONEOS_ARCHS armv7;armv7s;arm64)
+set(IPHONESIMULATOR_ARCHS i386;x86_64)
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/library/std/libcxx.cmake")
```
ruslo: Do we really need this one? I think libc++ used by default in iOS Xcode projects?

scripts/Info.plist
```diff
@@ -9,7 +9,7 @@
     <key>CFBundleInfoDictionaryVersion</key>
     <string>6.0</string>
     <key>CFBundlePackageType</key>
-    <string>dSYM</string>
+    <string>FMWK</string>
```
ruslo: One more reason to implement this in CMake. Usually this file can be tuned by user's variables. Just thought aloud.

headupinclouds:
I just realized this was never merged.  Shall I submit a separate pull request for a default of: `<string>FMWK</string>`.  Without this the dynamic frameworks would crash when called by another framework in an iOS project.

I'm also adding the create_framework.py mods to allow user specified plist files, but I think it would be a good idea to have this fix in place.

headupinclouds:
resubmitted just the FMWK mod in pull request #41.  closing this one, which has conflicts.

PR #30, opened 2015-07-08
Raspberry Pi 2 Toolchain Support

pretyman:
I added a generic gcc-based cross compiler support and implemented the RaspberryPi 2 Toolchain support on top of it.

Tested under Linux, later I will get a change to test it under MacOS 10.9.x

### DIFF
--- a/compiler/gcc-cross-compile.cmake
+++ b/compiler/gcc-cross-compile.cmake
@@ -23,18 +23,18 @@ endif()
 string(COMPARE EQUAL
     "${CROSS_COMPILE_TOOLCHAIN_PATH}"
     ""
-    _is_empty_cross_compile_toolchain_path
+    _is_empty
 )
-if(_is_empty_cross_compile_toolchain_path)
+if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PATH not set. Set it to the absolute path of the \"bin\" directory for the toolchain")
 endif()
 
 string(COMPARE EQUAL
     "${CROSS_COMPILE_TOOLCHAIN_PREFIX}"
     ""
-    _is_empty_cross_compile_toolchain_prefix
+    _is_empty
 )
-if(_is_empty_cross_compile_toolchain_prefix)
+if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PREFIX not set. Set it to the triplet of the toolchain")
 endif()
 
@@ -45,11 +45,11 @@ if (_has_sysroot)
   set(SYSROOT_COMPILE_FLAG "--sysroot=${CROSS_COMPILE_SYSROOT}")
   polly_add_cache_flag(
       CMAKE_C_FLAGS
-      "${CMAKE_C_FLAGS} ${SYSROOT_COMPILE_FLAG}"
+      "${SYSROOT_COMPILE_FLAG}"
   )
   polly_add_cache_flag(
       CMAKE_CXX_FLAGS
-      "${CMAKE_CXX_FLAGS} ${SYSROOT_COMPILE_FLAG}"
+      "${SYSROOT_COMPILE_FLAG}"
   )
 endif()
 
--- a/compiler/gcc-cross-compile.cmake
+++ b/compiler/gcc-cross-compile.cmake
@@ -6,7 +6,6 @@
 #
 #  Variables:
 #   CROSS_COMPILE_TOOLCHAIN_PATH=/path/to/toolchain/bin
-#     
 #   CROSS_COMPILE_TOOLCHAIN_PREFIX=arm-unknown-linux-gnueabihf
 #   CROSS_COMPILE_SYSROOT=/path/to/sysroot [optional]
 # ------------------------------------------------------------------------------
@@ -32,7 +31,7 @@ if (_has_sysroot)
 endif()
 
 # The (...)_PREFIX variable name refers to the Cross Compiler Triplet
-#set(TOOLCHAIN_PATH_AND_PREFIX ${CROSS_COMPILE_PATH}/${CROSS_COMPILE_TOOLCHAIN_TRIPLET})
+set(TOOLCHAIN_PATH_AND_PREFIX ${CROSS_COMPILE_TOOLCHAIN_PATH}/${CROSS_COMPILE_TOOLCHAIN_PREFIX})
 set(CMAKE_C_COMPILER     "${TOOLCHAIN_PATH_AND_PREFIX}-gcc"     CACHE PATH "C compiler" )
 set(CMAKE_CXX_COMPILER   "${TOOLCHAIN_PATH_AND_PREFIX}-g++"     CACHE PATH "C++ compiler" )
 set(CMAKE_ASM_COMPILER   "${TOOLCHAIN_PATH_AND_PREFIX}-as"      CACHE PATH "Assembler" )
@@ -49,24 +48,3 @@ set(CMAKE_NM             "${TOOLCHAIN_PATH_AND_PREFIX}-nm"      CACHE PATH "nm" 
 set(CMAKE_OBJCOPY        "${TOOLCHAIN_PATH_AND_PREFIX}-objcopy" CACHE PATH "objcopy" )
 set(CMAKE_OBJDUMP        "${TOOLCHAIN_PATH_AND_PREFIX}-objdump" CACHE PATH "objdump" )
 set(CMAKE_RANLIB         "${TOOLCHAIN_PATH_AND_PREFIX}-ranlib"  CACHE PATH "ranlib" )
-
-#set( CMAKE_C_PLATFORM_ID Linux )
-#set( CMAKE_C_COMPILER_ID GNU)
-#set( CMAKE_C_SIZEOF_DATA_PTR 4 )
-#set( CMAKE_C_HAS_ISYSROOT 1 )
-#set( CMAKE_C_COMPILER_ABI ELF )
-#set( CMAKE_CXX_COMPILER_ID GNU)
-#set( CMAKE_CXX_PLATFORM_ID Linux)
-#set( CMAKE_CXX_SIZEOF_DATA_PTR 4)
-#set( CMAKE_CXX_HAS_ISYSROOT 1)
-#set( CMAKE_CXX_COMPILER_ABI ELF)
-#set( CMAKE_CXX_SOURCE_FILE_EXTENSIONS cc cp cxx cpp CPP c++ C )
-## force ASM compiler (required for CMake < 2.8.5)
-#set( CMAKE_ASM_COMPILER_ID_RUN TRUE )
-#set( CMAKE_ASM_COMPILER_ID GNU )
-#set( CMAKE_ASM_COMPILER_WORKS TRUE )
-#set( CMAKE_ASM_COMPILER_FORCED TRUE )
-#set( CMAKE_COMPILER_IS_GNUASM 1)
-#set( CMAKE_ASM_SOURCE_FILE_EXTENSIONS s S asm )
-
-
--- a/compiler/gcc-cross-compile.cmake
+++ b/compiler/gcc-cross-compile.cmake
@@ -13,21 +13,21 @@
 cmake_minimum_required( VERSION 2.8 )
 
 if(DEFINED POLLY_COMPILER_GCC_CROSS_COMPILE)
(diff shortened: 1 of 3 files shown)


compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,50 @@
...
+#  Variables:
+#   CROSS_COMPILE_TOOLCHAIN_PATH=/path/to/toolchain/bin
+#   CROSS_COMPILE_TOOLCHAIN_PREFIX=arm-unknown-linux-gnueabihf
+#   CROSS_COMPILE_SYSROOT=/path/to/sysroot [optional]
+# ------------------------------------------------------------------------------
+
+cmake_minimum_required( VERSION 2.8 )
+
+if(DEFINED POLLY_COMPILER_GCC_CROSS_COMPILE)
+	return()
```
ruslo: Please remove tabs - use 2 spaces.

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2015 Alexandre Pretyman
+# All rights reserved.
+#
+# ------------------------------------------------------------------------------
+#  GCC Based Cross Compiler CMake toolchain file
+#
+#  Variables:
+#   CROSS_COMPILE_TOOLCHAIN_PATH=/path/to/toolchain/bin
+#   CROSS_COMPILE_TOOLCHAIN_PREFIX=arm-unknown-linux-gnueabihf
+#   CROSS_COMPILE_SYSROOT=/path/to/sysroot [optional]
```
ruslo: I guess this is a CMake variables that must be set in toolchain file from the top. Probably we need to check they are not empty and report descriptive error message otherwise.

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,50 @@
...
+
+if( CMAKE_TOOLCHAIN_FILE )
+ # touch toolchain variable to suppress "unused variable" warning
+endif()
+
+#Check for if sysroot exists, optional, since it could be hardcoded in the compiler
+string(COMPARE NOTEQUAL "${CROSS_COMPILE_SYSROOT}" "" _has_sysroot)
+if (_has_sysroot)
+	set(SYSROOT_COMPILE_FLAG "--sysroot=${CROSS_COMPILE_SYSROOT}")
+	set(CMAKE_C_FLAGS        "${CMAKE_C_FLAGS} ${SYSROOT_COMPILE_FLAG}")
```
ruslo: This one must be set to cache. You can use `polly_add_cache_flag` function. See for example: https://github.com/ruslo/polly/blob/master/flags/cxx11.cmake

compiler/gcc-cross-compile.cmake
pretyman: Ok, for now I will do a manual check. I also have to target RaspberryPi 1, so when I get around to do its toolchain, I will create a `poly_test_string_not_empty` with a way to add an error message and refactor this file

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,73 @@
...
+else()
+  set(POLLY_COMPILER_GCC_CROSS_COMPILE TRUE)
+endif()
+
+if( CMAKE_TOOLCHAIN_FILE )
+  # touch toolchain variable to suppress "unused variable" warning
+endif()
+
+string(COMPARE EQUAL
+    "${CROSS_COMPILE_TOOLCHAIN_PATH}"
```
ruslo: Please remove traling whitespaces, thanks.

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,73 @@
...
+    _is_empty_cross_compile_toolchain_path
+)
+if(_is_empty_cross_compile_toolchain_path)
+  poly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PATH not set. Set it to the absolute path of the \"bin\" directory for the toolchain")
+endif()
+
+string(COMPARE EQUAL
+    "${CROSS_COMPILE_TOOLCHAIN_PREFIX}" 
+    "" 
+    _is_empty_cross_compile_toolchain_prefix
```
ruslo: Since this variable used only once I guess it can be just `_is_empty` (?)

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,73 @@
...
+endif()
+
+
+#Check for if sysroot exists, optional, since it could be hardcoded in the compiler
+string(COMPARE NOTEQUAL "${CROSS_COMPILE_SYSROOT}" "" _has_sysroot)
+if (_has_sysroot)
+  set(SYSROOT_COMPILE_FLAG "--sysroot=${CROSS_COMPILE_SYSROOT}")
+  polly_add_cache_flag(
+      CMAKE_C_FLAGS   
+      "${CMAKE_C_FLAGS} ${SYSROOT_COMPILE_FLAG}"
```
ruslo: No need to use `CMAKE_C_FLAGS` in `"${CMAKE_C_FLAGS} ${SYSROOT_COMPILE_FLAG}"`

raspberrypi2-cxx11.cmake
```diff
@@ -0,0 +1,49 @@
...
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_clear_environment_variables.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_fatal_error.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_add_cache_flag.cmake")
+
+polly_init(
+    "RaspberryPi 2 Cross Compile / C++11"
+    "Unix Makefiles"
+)
+set(RAPSBERRYPI2_COMPILE_FLAGS
```
ruslo: Trailing whitespaces

compiler/gcc-cross-compile.cmake
pretyman: Do you use a Tool to spot them? How do you detect them?

pretyman: `grep -r '[[:blank:]]$' .` was one google away :)

ruslo: I have a GVim settings with highlights, also `git diff` show them.

raspberrypi2-cxx11.cmake
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2015, Alexandre Pretyman
+# All rights reserved.
+
+if(DEFINED POLLY_RASPBERRYPI2_CXX11_CMAKE)
+  return()
+else()
+  set(POLLY_RASPBERRYPI2_CXX11_CMAKE 1)
+endif()
+
+#Maybe this line should be in polly_init and not in polly_common?
```
ruslo: For now `polly_init` set variables like `POLLY_TOOLCHAIN_TAG` and `POLLY_TOOLCHAIN_NAME` and `polly_common`  check that this variables are set but I guess it can be improved in future. I'm planning to rethink/refactor a lot of stuff (like https://github.com/ruslo/polly/issues/31) and will remove this one too.

raspberrypi2-cxx11.cmake
```diff
@@ -0,0 +1,54 @@
...
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/utilities")
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
```
ruslo: For for in toolchains first we include `polly_init` explicitly, then use it, then include `polly_common` explicitly, then other modules available for include.

raspberrypi2-cxx11.cmake
```diff
@@ -0,0 +1,54 @@
...
+# All rights reserved.
+
+if(DEFINED POLLY_RASPBERRYPI2_CXX11_CMAKE)
+  return()
+else()
+  set(POLLY_RASPBERRYPI2_CXX11_CMAKE 1)
+endif()
+
+#Maybe this line should be in polly_init and not in polly_common?
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/utilities")
```
ruslo: Not needed if you include `polly_common` explicitly

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,73 @@
...
+    ""
+    _is_empty
+)
+if(_is_empty)
+  polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PREFIX not set. Set it to the triplet of the toolchain")
+endif()
+
+
+#Check for if sysroot exists, optional, since it could be hardcoded in the compiler
+string(COMPARE NOTEQUAL "${CROSS_COMPILE_SYSROOT}" "" _has_sysroot)
```
ruslo: In our particular case I think this variable must be set always (?)

compiler/gcc-cross-compile.cmake
```diff
@@ -0,0 +1,73 @@
...
+  # touch toolchain variable to suppress "unused variable" warning
+endif()
+
+string(COMPARE EQUAL
+    "${CROSS_COMPILE_TOOLCHAIN_PATH}"
+    ""
+    _is_empty
+)
+if(_is_empty)
+  polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PATH not set. Set it to the absolute path of the \"bin\" directory for the toolchain")
```
ruslo: I think this message is not quite helpful since user need to set environment variable `RASPBERRYPI_CROSS_COMPILE_TOOLCHAIN_PATH` (?)

pretyman: Agreed, I should just notify that variable is not set and move this message to `raspberrypi2-cxx11.cmake`

ruslo: I guess we should check `CROSS_COMPILE_TOOLCHAIN_PATH` CMake variable in `compiler/gcc-cross-compile.cmake` and report `CROSS_COMPILE_TOOLCHAIN_PATH not set` (internal error) and check `RASPBERRYPI_CROSS_COMPILE_TOOLCHAIN_PATH` environment variable in `raspberrypi2-cxx11.cmake` and report `RASPBERRYPI_* environment variable not set, please set it to ...` (user error).

ruslo: @pretyman Also just for your information I see no updates if you just push something new to this pull request. Feel free to write to me here or by email.

compiler/gcc-cross-compile.cmake
pretyman: hum... well, strictly speaking it is not fully required, but it is a very good idea to have it set, as when compiling Qt for the Raspberry Pi, it seems to be needed. For now I guess you are right and it is best to make it required, because having it required would make things just work, leaving it optional could make the toolchain not work.

raspberrypi2-cxx11.cmake
pretyman: Ok

compiler/gcc-cross-compile.cmake
pretyman: Thanks, I didn't know that. I just pushed the changes :-)

ruslo:
@pretyman Thanks! By the way have you tried it with Hunter?

ruslo:
@pretyman Wiki updated: https://github.com/ruslo/polly/wiki/Toolchain-list#raspberrypi2-cxx11

pretyman:
@ruslo Yes, it works with hunter, as the output of hunter-simple example (which uses GTest)

```
Running main() from gtest_main.cc
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from Simple
[ RUN      ] Simple.simple
[       OK ] Simple.simple (0 ms)
[----------] 1 test from Simple (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 1 test.
```
:-)

ruslo:
@pretyman Awesome, thanks! :) I've just added a note: https://github.com/ruslo/hunter/commit/5cf367f7099d4fa17e9540c646063c9ce16a9c36

pretyman:
@ruslo: FYI: Autotools needs fixing though, I shall be working on that soon

PR #33, opened 2015-07-21
Add an option to generate xml output from ctest

oliverdaniell:
This adds an extra option to build.py that saves the ctest output to an xml file that can be passed by the xunit plugin in jenkins.

### DIFF
--- a/bin/build.py
+++ b/bin/build.py
@@ -60,6 +60,8 @@ parser.add_argument(
 )
 
 parser.add_argument('--test', action='store_true', help="Run ctest after build")
+parser.add_argument('--test_xml', help="Save ctest output to xml")
+
 parser.add_argument('--pack', action='store_true', help="Run cpack after build")
 parser.add_argument(
     '--nobuild', action='store_true', help="Do not build (only generate)"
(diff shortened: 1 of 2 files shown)


bin/build.py
```diff
@@ -60,6 +60,8 @@
 )
 
 parser.add_argument('--test', action='store_true', help="Run ctest after build")
+parser.add_argument('--test_xml', help="Save ctest output to xml")
```
ruslo: All toolchain's name use `-` instead of `_` (also it's `--framework-device`). I think we can use same conventions here: `--test-xml`.

bin/build.py
```diff
@@ -302,7 +304,7 @@
   os.chdir(build_dir)
   if args.test:
```
ruslo: You can test `if args.test or args.test_xml` so there will be no need to add `--test` every time you use `--test-xml`.

oliverdaniell:
Changes look good, do I need to do anything on my end?

ruslo:
> Changes look good, do I need to do anything on my end?

> All toolchain's name use - instead of _ (also it's --framework-device). I think we can use same conventions here: --test-xml

> You can test if args.test or args.test_xml so there will be no need to add --test every time you use --test-xml

It would be nice if you'll update this pull request with this fixes. You can just push to your branch `master` (with rebase probably).

oliverdaniell:
changes applied

PR #35, opened 2015-07-30
Add "--strip" option to trigger install/strip target for CMake Unix Makefile generator

headupinclouds:
…trip target

Specifying --strip will imply the local directory --install behavior if not specified

### DIFF
--- a/bin/build.py
+++ b/bin/build.py
@@ -82,6 +82,9 @@ parser.add_argument(
     help="Create framework for device (exclude simulator architectures)"
 )
 parser.add_argument(
+    '--strip', action='store_true', help="Run strip/install cmake targets"
+)
+parser.add_argument(
     '--clear',
     action='store_true',
     help="Remove build and install dirs before build"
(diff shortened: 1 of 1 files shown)


ruslo:
@headupinclouds Hi David. Does it actually works? :) I've tried simple example and I see no difference between `install` and `install/strip`:

```cmake
# CMakeLists.txt
cmake_minimum_required(VERSION 3.0)
project(Foo)

add_library(foo foo.cpp)
install(TARGETS foo DESTINATION lib)
```

```bash
> build.py --toolchain libcxx --clear --verbose --config Debug --install
...
/.../cmake -P cmake_install.cmake
> ls -la _install/libcxx/lib/libfood.a
... 71800 ... _install/libcxx/lib/libfood.a
> strings _install/libcxx/lib/libfood.a > nostrip.strings
```

```bash
> build.py --toolchain libcxx --clear --verbose --config Debug --strip
...
/.../cmake -DCMAKE_INSTALL_DO_STRIP=1 -P cmake_install.cmake
> ls -la _install/libcxx/lib/libfood.a
... 71800 ... _install/libcxx/lib/libfood.a
> strings _install/libcxx/lib/libfood.a > strip.strings
```

As you can see sizes are quite the same and *.strings files are too:
```bash
> diff nostrip.strings strip.strings
> echo $?
0
```

Meanwhile:
```bash
> strip -S _install/libcxx/lib/libfood.a -o libfoo.stripped
> ls -la libfoo.stripped
... 14056 ... libfoo.stripped
```

By the way I see no variable `CMAKE_INSTALL_DO_STRIP` used in `cmake_install.cmake` script.

headupinclouds:
I've noticed strip does get called, but I think C++ code with RTTI and/or
exceptions can prevent symbols from being stripped.  I believe templates
can also cause problems in some cases, so this may not be all that useful
in practice.

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/strip:
symbols referenced by indirect symbol table entries that can't be stripped
in: example.0.1.0.dylib


I assume there are simpler (pure C) cases where it will work, but polly
does seem to be calling the install/strip target properly.  I can try to
test a C example.

ruslo:
@headupinclouds I've tested it on executable and it works.

PR #41, opened 2015-07-31
Use FMWK  for CFBundlePackageType

headupinclouds:


headupinclouds:
resubmit FMWK mod from #28.

PR #42, opened 2015-08-03
unique cmake include guards (copy and paste error)

headupinclouds:


ruslo:
@headupinclouds Thanks

PR #58, opened 2015-12-09
targets mips/mips64 - api lvl 21 - ndk r10e

skypjack:
As in the title, added support for targets mips/mips64, api level 21, ndk r10e.


ruslo:
Thank you. Applied with minor fix: https://github.com/ruslo/polly/commit/fbec89ee36a559418238e8426b2c280bf78c2fbc

In fact it will not produce any errors, just keep it constistent.

PR #65, opened 2016-02-05
Pr.gcc.hid.fpic

headupinclouds:


ruslo:
Since flag set explicitly there is no need to use `CMAKE_POSITION_INDEPENDENT_CODE`: https://github.com/ruslo/polly/commit/b74b95a812351147081adb4170599c1a94f4cb33

PR #71, opened 2016-03-18
Log to `log.txt`. When already exists rename the old `log.txt` to `log-<n>.txt`.

xpol:
Log to `log.txt`. When already exists rename the old `log.txt` to `log-<n>.txt`.

This is good for checking the log: `cat _logs/polly/log.txt` will always show the last log.
I don't needs to copy paste or edit my last command in bash.


ruslo:
> Log to log.txt. When already exists rename the old log.txt to log-<n>.txt

Make sense. Thanks!

PR #75, opened 2016-04-14
Rename to polly add shell scripts

xpol:
I'm new to Python/shell/batchfile, please review first.


bin/polly
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+/usr/bin/env python3 `dirname $0`/polly.py "$@"
```
ruslo: Will not work if `polly` called through symbolic link. Should be like

```
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```

See [SO: Can a Bash script tell what directory it's stored in?](http://stackoverflow.com/a/246128/2288008)

bin/polly.bat
```diff
@@ -0,0 +1 @@
+python3 %~dp0\\polly.py %*
```
ruslo: On windows executable called simply `python.exe`, not `python3.exe`.

xpol:
@ruslo Is this OK to merge?

ruslo:
> @ruslo Is this OK to merge?

@xpol Thanks for the updates, will take a look. Just for your information when you push new commits to pull requests I'm not receiving any notifications. Please add some messages next time, see also:
* http://0.readthedocs.org/en/latest/git.html#notify

xpol:
@ruslo 
Got that! Thank you very much!

PR #78, opened 2016-04-18
WIP: support for ndk r11c

skypjack:
The aim is to migrate to the latest ndk (that is r11c, available since March 2016).
_.travis.yml_ and _bin/install-ci-dependencies.py_ have still to be updated, but first of all a packed version of the r11c ndk should be added to the repo _https://github.com/hunter-packages/android-ndk_, right?

PR #79, opened 2016-04-18
moving to the ndk r11c

skypjack:
see https://github.com/ruslo/polly/pull/78

### DIFF
--- a/bin/install-ci-dependencies.py
+++ b/bin/install-ci-dependencies.py
@@ -193,9 +193,9 @@ def get_android_url():
         return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r10e-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Darwin.tar.gz', 'e568e9a8f562e7d1bc06f93e6f7cc7f44df3ded2'
     if os.getenv('TOOLCHAIN') == 'android-ndk-r11c-api-19-armeabi-v7a-neon':
       if platform.system() == 'Linux':
-        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Linux.tar.gz', 'b90d03d11cc1c5770e7851924a60e9819b578960'
+        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.1/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Linux.tar.gz', '2e0da01961e0031bfd7d8db6ce4a15372bd8c3e8'
       if platform.system() == 'Darwin':
-        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Darwin.tar.gz', '07f2425fa99377a678949314330ec7e5ebc597f8'
+        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.1/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Darwin.tar.gz', '664b3c8104142de2af16f887c19d1b2e618725cb'
   return get_android_full_version_url()
 
 if is_android:
--- a/bin/install-ci-dependencies.py
+++ b/bin/install-ci-dependencies.py
@@ -172,12 +172,17 @@ else:
   sys.exit('Unknown system: {}'.format(platform.system()))
 
 def get_android_full_version_url():
-  if platform.system() == 'Darwin':
-    return 'http://dl.google.com/android/repository/android-ndk-r11c-darwin-x86_64.zip', '4ce8e7ed8dfe08c5fe58aedf7f46be2a97564696',
-  elif platform.system() == 'Linux':
-    return 'http://dl.google.com/android/repository/android-ndk-r11c-linux-x86_64.zip', 'de5ce9bddeee16fb6af2b9117e9566352aa7e279',
-  else:
-    sys.exit('Android supported only for Linux and OSX')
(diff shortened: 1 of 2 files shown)


bin/install-ci-dependencies.py
```diff
@@ -173,15 +173,15 @@ def unpack(self):
 
 def get_android_url():
   if os.getenv('TRAVIS'):
-    if os.getenv('TOOLCHAIN') == 'android-ndk-r10e-api-19-armeabi-v7a-neon':
```
ruslo: Both NDK versions can be supported simply by not removing old code, just keep checking toolchain name

skypjack:
Just seen the comment on #78.
Let me update the commit, so as to support both the `ndk`s.

bin/install-ci-dependencies.py
skypjack: Just seen the comment on #78.
Let me update the commit, so as to support both the `ndk`s.

skypjack: Ok, updated. Please, check it for I'm not sure about that. Thank you.

bin/install-ci-dependencies.py
```diff
@@ -175,15 +175,27 @@ def get_android_url():
   if os.getenv('TRAVIS'):
     if os.getenv('TOOLCHAIN') == 'android-ndk-r10e-api-19-armeabi-v7a-neon':
       if platform.system() == 'Linux':
-        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r10e-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Linux.tar.gz', '847177799b0fe4f7480f910bbf1815c3e3fed0da'
-      if platform.system() == 'Darwin':
+        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r10e-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Linux.tar.gz', '847177799b0fe4f7480f910bbf1815c3e3fed0da'      if platform.system() == 'Darwin':
```
ruslo: I think there is a syntax error in this line

bin/install-ci-dependencies.py
```diff
@@ -175,15 +175,27 @@ def get_android_url():
...
-  elif platform.system() == 'Linux':
-    return 'http://dl.google.com/android/ndk/android-ndk-r10e-linux-x86_64.bin', 'c685e5f106f8daa9b5449d0a4f21ee8c0afcb2f6',
-  else:
-    sys.exit('Android supported only for Linux and OSX')
+    if os.getenv('TOOLCHAIN') == 'android-ndk-r11c-api-19-armeabi-v7a-neon':
+      if platform.system() == 'Linux':
+        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Linux.tar.gz', 'b90d03d11cc1c5770e7851924a60e9819b578960'
+      if platform.system() == 'Darwin':
+        return 'https://github.com/hunter-packages/android-ndk/releases/download/v1.0.0/android-ndk-r11c-arm-linux-androideabi-4.9-gnu-libstdc.-4.9-armeabi-v7a-android-19-arch-arm-Darwin.tar.gz', '07f2425fa99377a678949314330ec7e5ebc597f8'
+  if os.getenv('TOOLCHAIN') == 'android-ndk-r10e-api-19-armeabi-v7a-neon':
```
ruslo: Should be under `os.getenv('TRAVIS')` condition. Logic is next:
```
if(TRAVIS)
  if(toolchain-name is good) # i.e. have pruned version
    return pruned-version
  else
    return full-version
else
  # always return full version for non-ci environment
  return full-version
```
feel free to refactor

skypjack: As full version I'd use the upstream, http://dl.google.com/android/repository/android-ndk-r11c-darwin-x86_64.zip and http://dl.google.com/android/repository/android-ndk-r11c-linux-x86_64.zip. Is it fine?

ruslo: > Is it fine?

Sure. What is the alternative?

skypjack: r10e? :-)
It's still available for the download.

bin/install-ci-dependencies.py
skypjack: What's the syntax error? I can't see it, sorry. I'm far from being a py developer and I'm doing my best.

bin/install-ci-dependencies.py
ruslo: > r10e? 

Full `r11c` for toolchain with name `android-ndk-r11c-*`, full `r10e` for toolchain with name `android-ndk-r10e-*`, error if toolchain `android-*` and doesn't match previous names, nothing if toolchain doesn't match `android-*`.

bin/install-ci-dependencies.py
ruslo:
```python
  return '...', '...'      if platform.system() == 'Darwin':
```

Should be
```python
  return '...', '...'
if platform.system() == 'Darwin':
```

skypjack: Damn it, completely missed, sorry. Fixed.

bin/install-ci-dependencies.py
skypjack: Ok, added full version for both the toolchains and error if the toolchain name does not match. Is it as you thought it?

ruslo: Thanks, looks good for me. Let's wait for testing now.

skypjack: Failed indeed. What's going on?
I'm not sure about what failed, it seems to be due to a problem of permissions.
Let me know if I can help you solving the problem.

ruslo: > I'm not sure about what failed, it seems to be due to a problem of permissions

Yes, problem is in `*.tar.gz` pruned archive. There is no executable permissions. Actually it's a Python bug: https://bugs.python.org/issue15795. Working on workaround now.

ruslo: @skypjack Please try this release (need to update URL and SHA1):
* https://github.com/hunter-packages/android-ndk/releases/tag/v1.0.1

ruslo: Just for the record this bug not affect `r10e` version because full original archive have format `.bin` instead of `.zip`.

skypjack: Push just done, updated urls and digests.

Issue #81, opened 2016-06-03
Verbosity levels


yowidin:
Currently the polly script either compiles everything silently or prints way too much information.

With the `--verbose` flag the `generate_command.append('-DCMAKE_VERBOSE_MAKEFILE=ON')` command within the polly.py script makes build output almost unreadable: it is very hard to notice any compiler warnings and errors while building. Of course you could simply pass a `--fwd CMAKE_VERBOSE_MAKEFILE=OFF` argument to the polly but I personally don't think that verbose output should be default behavior.

Without the `--verbose` flag where is literally no build progress at all.

My proposal is to introduce something like verbosity levels:
- silent - same as current behavior without the `--verbose` flag
- normal - normal CMake output without compiler arguments `-DCMAKE_VERBOSE_MAKEFILE=OFF`
- full - same as current behavior with the `--verbose` flag

ruslo:
> My proposal is to introduce something like verbosity levels

seems good for me, feel free to send pull request

PR #82, opened 2016-06-05
Implemented verbosity levels

yowidin:
#81 Implementation

There are three verbosity levels now:
- silent - only basic information is shown (e.g. python version, build
  duration, logs path)
- **normal** in addition to the silent level CMake output is also sown
- full - verbose output from Makefile builds is also shown

The levels above are controlled with the `--verbosity-level` flag,
the `--verbose` flag is mutually exclusive with it and sets the
verbosity level to `full`

### DIFF
--- a/bin/detail/call.py
+++ b/bin/detail/call.py
@@ -46,7 +46,7 @@ def teed_call(cmd_args, logging):
   )
   threads = []
 
-  if logging.verbose:
+  if logging.verbosity != 'silent':
     threads.append(tee(p.stdout, logging.discard, logging.log_file, sys.stdout))
     threads.append(tee(p.stderr, logging.discard, logging.log_file, sys.stderr))
   else:
@@ -63,7 +63,8 @@ def call(call_args, logging, cache_file='', ignore=False, sleep=0):
   for i in call_args:
     pretty += '  `{}`\n'.format(i)
   pretty += ']\n'
-  print(pretty)
+  if logging.verbosity == 'full':
+    print(pretty)
   logging.log_file.write(pretty)
 
   # print one line version
@@ -71,7 +72,7 @@ def call(call_args, logging, cache_file='', ignore=False, sleep=0):
   for i in call_args:
     oneline += ' "{}"'.format(i)
   oneline = "[{}]>{}\n".format(os.getcwd(), oneline)
-  if logging.verbose:
+  if logging.verbosity == 'full':
     print(oneline)
   logging.log_file.write(oneline)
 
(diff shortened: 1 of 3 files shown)


bin/detail/call.py
```diff
@@ -63,15 +63,16 @@ def call(call_args, logging, cache_file='', ignore=False, sleep=0):
   for i in call_args:
     pretty += '  `{}`\n'.format(i)
   pretty += ']\n'
-  print(pretty)
+  if logging.verbosity == 'full':
```
ruslo: > silent - same as current behavior without the --verbose flag

so it still should be printed always

bin/detail/call.py
```diff
@@ -63,15 +63,16 @@ def call(call_args, logging, cache_file='', ignore=False, sleep=0):
...
+    print(pretty)
   logging.log_file.write(pretty)
 
   # print one line version
   oneline = ''
   for i in call_args:
     oneline += ' "{}"'.format(i)
   oneline = "[{}]>{}\n".format(os.getcwd(), oneline)
-  if logging.verbose:
+  if logging.verbosity == 'full':
```
ruslo: I think it should be `verbosity != 'silent'` (?)

yowidin: Is this information really required in the `normal` case? For example as a developer you either checked out a project and want to build it, or just doing your routine  rebuilds, are the command invocation details really helpful in this case?

bin/detail/call.py
yowidin: Same as below - I picked an option that provides most (but not more) information required for the developer in day-to-day scenario leaving full output for either `--verbose` or `--verbose-level=full` cases.  Of course a can fix it and make the implementation follow the issue literally.

bin/detail/call.py
ruslo: `oneline` may be helpful in case you want to repeat the steps without Polly (e.g. file a bug without introducing Polly extra layer). In fact I don't use it very often, so I'm okay with `verbosity == 'full'`. However printing CMake commands is a good sanity checking (see line 66 comment).

bin/detail/call.py
ruslo: Just tried it. Looks kind of strange to me, I see some CMake configuration output, then some compile commands but don't see what command originally executed.

yowidin: OK. I guess I had a wrong approach to it. First I will fix issues on both lines and rethink what exactly would be a better solution (a theme for another issue).

yowidin:
@ruslo Fixed the normal level level

ruslo:
> Fixed the normal level

looks good, let's wait for testing finish just to be sure

PR #84, opened 2016-06-08
Added two new toolchains

skypjack:
Added toolchains `android-ndk-r11c-api-16-armeabi` and `android-ndk-r11c-api-16-armeabi-v7a`.


ruslo:
Thanks

PR #89, opened 2016-07-05
Debug verbosity for Polly and Hunter only shown on full in polly.py

s-brain:
Polly.py sets verbosity levels for Polly and Hunter to Debug all the time. This change reverts that

Looks like a mistaken indent removal


ruslo:
Yep, looks strange, was added here: https://github.com/ruslo/polly/commit/27f994716762f1d0ea5b69cc1ed1e5526f8f5794#diff-df17afae6a758d51ebefce8c1c045560L205

Thanks for the fix.

PR #93, opened 2016-08-08
Added emscripten Toolchain

pretyman:
Привет Ruslan!

I've added the C++ to JavaScript compiler [emscripten](https://github.com/kripken/emscripten)

emscripten provides internal implementations for `glfw` and `glew`. I wonder if I could somehow stub out the `hunter_add_package(glfw)` and `hunter_add_package(glew)` to make them not download/compile. 

Right now in my application I have:

```CMake
if(NOT EMSCRIPTEN)
    hunter_add_package(glfw)
endif()
```

But that doesn't scale to other packages. I saw that in hunter you determine the existence of an installed package with a `DONE` file. 

I wonder if I could mark these two as installed somehow.

### DIFF
--- a/compiler/emscripten.cmake
+++ b/compiler/emscripten.cmake
@@ -1,0 +1,21 @@
+# Copyright (c) 2016, Alexandre Pretyman
+# All rights reserved.
+
+if(DEFINED POLLY_COMPILER_EMSCRIPTEN_CMAKE)
+  return()
+else()
+  set(POLLY_COMPILER_EMSCRIPTEN_CMAKE 1)
+endif()
+
+include(polly_fatal_error)
+
+string(COMPARE EQUAL "$ENV{EMSCRIPTEN}" "" _is_empty)
+if(_is_empty)
+  polly_fatal_error(
+    "EMSCRIPTEN environment variable not set. Set up your Emscripten environment variables with emsdk_env.sh"
+  )
+endif()
+
+
+include("${CMAKE_CURRENT_LIST_DIR}/emscripten.toolchain.cmake")
+
--- a/compiler/emscripten.toolchain.cmake
+++ b/compiler/emscripten.toolchain.cmake
@@ -1,0 +1,272 @@
+# This file is a 'toolchain description file' for CMake.
+# It teaches CMake about the Emscripten compiler, so that CMake can generate makefiles
+# from CMakeLists.txt that invoke emcc.
+
+# To use this toolchain file with CMake, invoke CMake with the following command line parameters
+# cmake -DCMAKE_TOOLCHAIN_FILE=<EmscriptenRoot>/cmake/Modules/Platform/Emscripten.cmake
+#       -DCMAKE_BUILD_TYPE=<Debug|RelWithDebInfo|Release|MinSizeRel>
+#       -G "Unix Makefiles" (Linux and OSX)
+#       -G "MinGW Makefiles" (Windows)
+#       <path/to/CMakeLists.txt> # Note, pass in here ONLY the path to the file, not the filename 'CMakeLists.txt' itself.
+
+# After that, build the generated Makefile with the command 'make'. On Windows, you may download and use 'mingw32-make' instead.
+
+# The following variable describes the target OS we are building to.
+set(CMAKE_SYSTEM_NAME Emscripten)
+set(CMAKE_SYSTEM_VERSION 1)
+
+set(CMAKE_CROSSCOMPILING TRUE)
+
+# Advertise Emscripten as a 32-bit platform (as opposed to CMAKE_SYSTEM_PROCESSOR=x86_64 for 64-bit platform),
+# since some projects (e.g. OpenCV) use this to detect bitness.
+set(CMAKE_SYSTEM_PROCESSOR x86)
+
+# Tell CMake how it should instruct the compiler to generate multiple versions of an outputted .so library: e.g. "libfoo.so, libfoo.so.1, libfoo.so.1.4" etc.
+# This feature is activated if a shared library project has the property SOVERSION defined.
+set(CMAKE_SHARED_LIBRARY_SONAME_C_FLAG "-Wl,-soname,")
+
+# In CMake, CMAKE_HOST_WIN32 is set when we are cross-compiling from Win32 to Emscripten: http://www.cmake.org/cmake/help/v2.8.12/cmake.html#variable:CMAKE_HOST_WIN32
+# The variable WIN32 is set only when the target arch that will run the code will be WIN32, so unset WIN32 when cross-compiling.
+set(WIN32)
+
+# The same logic as above applies for APPLE and CMAKE_HOST_APPLE, so unset APPLE.
+set(APPLE)
+
+# And for UNIX and CMAKE_HOST_UNIX. However, Emscripten is often able to mimic being a Linux/Unix system, in which case a lot of existing CMakeLists.txt files can be configured for Emscripten while assuming UNIX build, so this is left enabled.
+set(UNIX 1)
+
+# Do a no-op access on the CMAKE_TOOLCHAIN_FILE variable so that CMake will not issue a warning on it being unused.
+if (CMAKE_TOOLCHAIN_FILE)
+endif()
+
+# In order for check_function_exists() detection to work, we must signal it to pass an additional flag, which causes the compilation
+# to abort if linking results in any undefined symbols. The CMake detection mechanism depends on the undefined symbol error to be raised.
+set(CMAKE_REQUIRED_FLAGS "-s ERROR_ON_UNDEFINED_SYMBOLS=1")
+
+# Locate where the Emscripten compiler resides in relative to this toolchain file.
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	get_filename_component(GUESS_EMSCRIPTEN_ROOT_PATH "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)
+	if (EXISTS "${GUESS_EMSCRIPTEN_ROOT_PATH}/emranlib")
+		set(EMSCRIPTEN_ROOT_PATH "${GUESS_EMSCRIPTEN_ROOT_PATH}")
+	endif()
+endif()
+
+# If not found by above search, locate using the EMSCRIPTEN environment variable.
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	set(EMSCRIPTEN_ROOT_PATH "$ENV{EMSCRIPTEN}")
+endif()
+
+# Abort if not found. 
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	message(FATAL_ERROR "Could not locate the Emscripten compiler toolchain directory! Either set the EMSCRIPTEN environment variable, or pass -DEMSCRIPTEN_ROOT_PATH=xxx to CMake to explicitly specify the location of the compiler!")
+endif()
+
+# Normalize, convert Windows backslashes to forward slashes or CMake will crash.
+get_filename_component(EMSCRIPTEN_ROOT_PATH "${EMSCRIPTEN_ROOT_PATH}" ABSOLUTE)
+
+if (NOT CMAKE_MODULE_PATH)
+	set(CMAKE_MODULE_PATH "")
+endif()
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${EMSCRIPTEN_ROOT_PATH}/cmake/Modules")
+
+set(CMAKE_FIND_ROOT_PATH "${EMSCRIPTEN_ROOT_PATH}/system")
+
+if (CMAKE_HOST_WIN32)
+	set(EMCC_SUFFIX ".bat")
+else()
+	set(EMCC_SUFFIX "")
+endif()
+
+# Specify the compilers to use for C and C++
+if ("${CMAKE_C_COMPILER}" STREQUAL "")
+	set(CMAKE_C_COMPILER "${EMSCRIPTEN_ROOT_PATH}/emcc${EMCC_SUFFIX}")
+endif()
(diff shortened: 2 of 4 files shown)


compiler/emscripten.cmake
```diff
@@ -0,0 +1,21 @@
...
+else()
+  set(POLLY_COMPILER_EMSCRIPTEN_CMAKE 1)
+endif()
+
+include(polly_fatal_error)
+
+string(COMPARE EQUAL "$ENV{EMSCRIPTEN}" "" _is_empty)
+if(_is_empty)
+  polly_fatal_error(
+    "EMSCRIPTEN environment variable not set. Set up your Emscripten environment variables with emsdk_env.sh"
```
ruslo: I guess you can load environment by `polly.py` by the way.

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+if (CMAKE_TOOLCHAIN_FILE)
+endif()
+
+# In order for check_function_exists() detection to work, we must signal it to pass an additional flag, which causes the compilation
+# to abort if linking results in any undefined symbols. The CMake detection mechanism depends on the undefined symbol error to be raised.
+set(CMAKE_REQUIRED_FLAGS "-s ERROR_ON_UNDEFINED_SYMBOLS=1")
+
+# Locate where the Emscripten compiler resides in relative to this toolchain file.
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	get_filename_component(GUESS_EMSCRIPTEN_ROOT_PATH "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)
```
ruslo: No tabs: http://0.readthedocs.io/en/latest/git.html#never-use-tabs

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+endif()
+
+# If not found by above search, locate using the EMSCRIPTEN environment variable.
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	set(EMSCRIPTEN_ROOT_PATH "$ENV{EMSCRIPTEN}")
+endif()
+
+# Abort if not found. 
+if ("${EMSCRIPTEN_ROOT_PATH}" STREQUAL "")
+	message(FATAL_ERROR "Could not locate the Emscripten compiler toolchain directory! Either set the EMSCRIPTEN environment variable, or pass -DEMSCRIPTEN_ROOT_PATH=xxx to CMake to explicitly specify the location of the compiler!")
```
ruslo: We can use `polly_fatal_error` here.

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+endif()
+
+# Normalize, convert Windows backslashes to forward slashes or CMake will crash.
+get_filename_component(EMSCRIPTEN_ROOT_PATH "${EMSCRIPTEN_ROOT_PATH}" ABSOLUTE)
+
+if (NOT CMAKE_MODULE_PATH)
```
ruslo: What this operation is about?

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+	set(CMAKE_MODULE_PATH "")
+endif()
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${EMSCRIPTEN_ROOT_PATH}/cmake/Modules")
```
ruslo: `list(APPEND ...)`

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+
+if (CMAKE_HOST_WIN32)
+	set(EMCC_SUFFIX ".bat")
+else()
+	set(EMCC_SUFFIX "")
+endif()
+
+# Specify the compilers to use for C and C++
+if ("${CMAKE_C_COMPILER}" STREQUAL "")
+	set(CMAKE_C_COMPILER "${EMSCRIPTEN_ROOT_PATH}/emcc${EMCC_SUFFIX}")
```
ruslo: Set it to cache with FORCE. No need to compare with empty string in this case.

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+endif()
+if ("${CMAKE_CXX_COMPILER}" STREQUAL "")
+	set(CMAKE_CXX_COMPILER "${EMSCRIPTEN_ROOT_PATH}/em++${EMCC_SUFFIX}")
+endif()
+
+if ("${CMAKE_AR}" STREQUAL "")
```
ruslo: No need to compare. If it's already in cache then nothing will happen.

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+# Therefore only CMAKE_FIND_ROOT_PATH should be used as a find directory. See http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_FIND_ROOT_PATH_MODE_INCLUDE.html
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+
+set(CMAKE_SYSTEM_INCLUDE_PATH "${EMSCRIPTEN_ROOT_PATH}/system/include")
+
+# We would prefer to specify a standard set of Clang+Emscripten-friendly common convention for suffix files, especially for CMake executable files,
+# but if these are adjusted, ${CMAKE_ROOT}/Modules/CheckIncludeFile.cmake will fail, since it depends on being able to compile output files with predefined names.
+#SET(CMAKE_LINK_LIBRARY_SUFFIX "")
```
ruslo: Can we remove this?

compiler/emscripten.toolchain.cmake
```diff
@@ -0,0 +1,272 @@
...
+set(CMAKE_EXE_LINKER_FLAGS_MINSIZEREL "-Os" CACHE STRING "Emscripten-overridden CMAKE_EXE_LINKER_FLAGS_MINSIZEREL")
+set(CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO "-O2 -g" CACHE STRING "Emscripten-overridden CMAKE_EXE_LINKER_FLAGS_RELWITHDEBINFO")
+set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "-O2" CACHE STRING "Emscripten-overridden CMAKE_SHARED_LINKER_FLAGS_RELEASE")
+set(CMAKE_SHARED_LINKER_FLAGS_MINSIZEREL "-Os" CACHE STRING "Emscripten-overridden CMAKE_SHARED_LINKER_FLAGS_MINSIZEREL")
+set(CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO "-O2 -g" CACHE STRING "Emscripten-overridden CMAKE_SHARED_LINKER_FLAGS_RELWITHDEBINFO")
+set(CMAKE_MODULE_LINKER_FLAGS_RELEASE "-O2" CACHE STRING "Emscripten-overridden CMAKE_MODULE_LINKER_FLAGS_RELEASE")
+set(CMAKE_MODULE_LINKER_FLAGS_MINSIZEREL "-Os" CACHE STRING "Emscripten-overridden CMAKE_MODULE_LINKER_FLAGS_MINSIZEREL")
+set(CMAKE_MODULE_LINKER_FLAGS_RELWITHDEBINFO "-O2 -g" CACHE STRING "Emscripten-overridden CMAKE_MODULE_LINKER_FLAGS_RELWITHDEBINFO")
+
+function(em_validate_asmjs_after_build target)
```
ruslo: If this function is not used in toolchain itself I think we should use separate CMake package like [Android-APK](https://github.com/ruslo/hunter/wiki/pkg.android.apk). We can develop them independently in this case.

ruslo:
> But that doesn't scale to other packages. I saw that in hunter you determine the existence of an installed package with a DONE file. 

I guess
```cmake
if(EMSCRIPTEN)
  return()
endif()
```
in `glfw/hunter.cmake` will do the trick.

pretyman:
Thanks Ruslan, I was hoping to do the less modifications of the original as possible so I could diff it easily. Or is there a way to recommend me controlling the patches I do in the original file. I don't mind doing the changes you suggested, I'm just thinking how to best control the evolution of  https://github.com/kripken/emscripten/blob/master/cmake/Modules/Platform/Emscripten.cmake compared to the one I am to add at polly.

The only thing that breaks the build in the original file is the `CMakeForceCompiler`, I tried stubbing it out my adding a CMakeForceCompiler.cmake with empty `CMAKE_FORCE_C_COMPILER` and `CMAKE_FORCE_CXX_COMPILER`, but no matter if I `list(APPEND CMAKE_MODULE_PATH ...)` or list(INSERT CMAKE_FORCE_C_COMPILER 0 ...)` I cannot override these macros.

If I could just get around this, there wouldn't be a need for this file to be in the `polly` repository and I could load it from `$ENV{EMSCRIPTEN}/cmake/.../Emscripten.cmake`.

I also tried undoing what is done by `CMAKE_FORCE_{C,CXX}_COMPILER` with:
```CMake
include("$ENV{EMSCRIPTEN}/cmake/Modules/Platforms/Emscripten.cmake")
set(CMAKE_C_COMPILER_ID_RUN)
set(CMAKE_C_COMPILER_FORCED)
set(CMAKE_C_COMPILER_ID)
set(CMAKE_CXX_COMPILER_ID_RUN)
set(CMAKE_CXX_COMPILER_FORCED)
set(CMAKE_CXX_COMPILER_ID)
```
But it still makes hunter fail the ABI test (can't understand why yet)

ruslo:
> Thanks Ruslan, I was hoping to do the less modifications of the original as possible so I could diff it easily

Okay, I see. It's reasonable, let's keep it as is then.

> The only thing that breaks the build in the original file is the CMakeForceCompiler

You can add an option `EMSCRIPTEN_FORCE_COMPILERS`, set it to `ON` by default and to `OFF` while loading through Polly. Sending it upstream will make everybody happy.

pretyman:
> You can add an option EMSCRIPTEN_FORCE_COMPILERS, set it to ON by default and to OFF while loading through Polly. Sending it upstream will make everybody happy.

Ok, I'll send the `emscripten` project the Find scripts for glew and glfw and propose this change, thanks!

compiler/emscripten.cmake
pretyman: if you are referring to use polly.py instead of the `emsdk_env.sh` script, I put this message here because `emsdk_env.sh` is the script that comes with emscripten to setup the environment variables needed by emscripten

ruslo: I mean that you can load this script from inside polly.py so you don't have to pollute your current space. Anyway it's not critical, just the direction how to make everything convenient.

pretyman:
Oh, it is a real shame that emscripten doesn't play nice with CMake:
  1. The CMake imported targets cannot have INTERFACE compile options [ Cannot specify compile options for imported target "glfw3" ]
  2. When the compile options that should be in the targets are put directly into the final app "binary" then this happens: https://cmake.org/Bug/view.php?id=15826 
  3. For some **COMPLETELY** unexplainable reason, if I `rm -rf _*` to remove polly's generated directories in the app's directory, and run `~/Development/polly/bin/polly --verbose --toolchain emscripten-cxx11` again, then hunter fails ABI test. When I make changes to the toolchain file, it gets back to normal (?)

Sometimes I just hate CMake, I will close this pull request and script my way through this one

ruslo:
> The CMake imported targets cannot have INTERFACE compile options [ Cannot specify compile options for imported target "glfw3" ]

What do you mean?

> When the compile options that should be in the targets are put directly into the final app "binary" then this happens: https://cmake.org/Bug/view.php?id=15826

Just for your info it's here now: https://gitlab.kitware.com/cmake/cmake/issues/15826. Well it's bug definitely, someone need to fix it.

> When I make changes to the toolchain file, it gets back to normal and pass the ABI test (wtf?)

Issues with cache, need to investigate. Also I don't know details of how emscripten compiler works, since it's not regular compiler I guess there might be unpredictable peculiarities.

ruslo:
> Sometimes I just hate CMake

The more I spent time in development in general the more I got feeling that every piece of sofware designed to torture my brain :) E.g. now I got 100+ broken Travis configs because installing of python.requests if screwed up.

pretyman:
> What do you mean?

Sorry, for not being explicit. Conventional CMake scripts use target `glfw3` to get information about the `glfw3` library, but in emscripten, from what I understood from the tests, to enable `glfw3` one has to pass the compilation flag `-s USE_GLFW=3`, so I wrongly thought about creating an imported target `glfw3` with `target_compile_options(glfw3 INTERFACE "-s;USE_GLFW=3")` but it gave me the error `Cannot specify compile options for imported target "glfw3"`. This was fixed by not using IMPORTED but using INTERFACE target, `Findglfw3.cmake` in emscripten toolchain:
```CMake
if(NOT TARGET glfw3)
  add_library(glfw3 INTERFACE)
  target_compile_options(glfw3 INTERFACE "-s;USE_GLFW=3")
endif()
```

Now in the CMakeLists.txt for my app I just keep

```
target_link_libraries(MyApp PUBLIC glfw3 ...)
```

And it works for Linux/Win/OSX/emscripten.
> Just for your info it's here now: https://gitlab.kitware.com/cmake/cmake/issues/15826. Well it's bug definitely, someone need to fix it.

Thanks, I've made a 1 minute very dirty patch for it here: https://github.com/pretyman/CMake/commit/c6ee8bfa03baf997f6e94c6a9a42de823ae6ee10. Needs to be reworked to go upstream.

Thanks for the insights, after sleeping over this problem the solutions came the next day. When I get everything nice and dandy, I'll open a pull request to polly again. Then we'll have emscripten in polly \o/

PR #94, opened 2016-08-11
Add ARM targets for VS 2012, 2013, 2015 and Win64 for VS 2012.

speps:
I hope the branch setup is okay.

I was missing the ARM targets for VS for UWP and Windows Store Apps. I noticed CMake also supported Win64 for VS 2012 which was missing too.

### DIFF
--- a/vs-11-2012-arm.cmake
+++ b/vs-11-2012-arm.cmake
@@ -1,0 +1,17 @@
+# Copyright (c) 2015, Ruslan Baratov
+# All rights reserved.
+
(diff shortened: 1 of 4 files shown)


ruslo:
Probably you also want to add them to toolchain table here: https://github.com/ruslo/polly/blob/5aaccb8ae4ac86ad3b67aa1d4e53db84ce3d557d/bin/detail/toolchain_table.py#L180 in case users want to use them via build.py. Feel free to open pull request with such patch. Thanks.

speps:
Done in #95.

PR #95, opened 2016-08-11
Update toolchain_table.py with new VS archs.

speps:
As requested :) What updates the wiki page by the way? Is that manual?

### DIFF
--- a/bin/detail/toolchain_table.py
+++ b/bin/detail/toolchain_table.py
@@ -168,19 +168,16 @@ if os.name == 'nt':
       Toolchain(
           'vs-11-2012-arm',
           'Visual Studio 11 2012 ARM',
-          arch='arm',
           vs_version='11'
       ),
       Toolchain(
           'vs-12-2013-arm',
           'Visual Studio 12 2013 ARM',
-          arch='arm',
           vs_version='12'
       ),
       Toolchain(
           'vs-14-2015-arm',
           'Visual Studio 14 2015 ARM',
-          arch='arm',
           vs_version='14'
       ),
       Toolchain(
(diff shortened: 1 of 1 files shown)


ruslo:
`arch` should be `x86` or `amd64`:

```bash
> build.py --toolchain vs-12-2013 --verbose --clear
Traceback (most recent call last):
  File "...\polly\bin\build.py", line 9, in <module>
    exec(open(join(dirname(abspath(__file__)), 'polly.py'), 'r').read())
  File "<string>", line 28, in <module>
  File "...\polly\bin\detail\toolchain_table.py", line 172, in <module>
    vs_version='11'
  File "...\polly\bin\detail\toolchain_table.py", line 34, in __init__
    self.verify()
  File "...\polly\bin\detail\toolchain_table.py", line 39, in verify
    assert(self.arch == 'amd64' or self.arch == 'x86')
AssertionError
```

This variable used for NMake generators to load environment by `vcvarsall.bat`:
* https://github.com/ruslo/polly/blob/31e59ed80f62fc7e93ecbee74472366454e28bcb/bin/detail/get_nmake_environment.py#L30

I don't know what we have to set for arm architecture. Since there is no NMake + arm toolchain I guess it's okay to set this variable to empty string just like MDD did:
* https://github.com/ruslo/polly/blob/31e59ed80f62fc7e93ecbee74472366454e28bcb/bin/detail/toolchain_table.py#L165

speps:
Default value is empty is I just removed the value.

ruslo:
> What updates the wiki page by the way? Is that manual?

Nope. I'm planning to refactor a lot of code, this includes automatic docs generation (at least partially), but I don't know when I will have time to do that. Also I want to move from wiki to readthedocs: http://polly.readthedocs.io/en/latest/

PR #97, opened 2016-09-01
Added emscripten Toolchain

pretyman:
Reopened the pull request from: https://github.com/ruslo/polly/pull/93

I'm not sure I understood your comments here: https://github.com/ruslo/polly/pull/93#discussion_r73873153 - isn't this script loaded by polly? I didn't really understand what you meant by " pollute your current space", if you were referring to the test to `$ENV{EMSCRIPTEN}`, I prefer to have in CMake and not in polly.py because of IDE integration

It works with original CMake without needed to be patched [ emscripten doesn't really suffer from: https://cmake.org/Bug/view.php?id=15826 as these are not compilation flags as described in the issue, but linker flags, however it seems to be a real bug elsewhere: https://gitlab.kitware.com/cmake/cmake/issues/15826 ]

Added package emulation into compiler/emscripten [ I guess in this case it is tied to the toolchain, so I would think it made sense to have these in polly ]

### DIFF
--- a/bin/detail/toolchain_table.py
+++ b/bin/detail/toolchain_table.py
@@ -91,6 +91,7 @@ toolchain_table = [
     Toolchain('android-ndk-r11c-api-21-x86-64-hid', 'Unix Makefiles'),
     Toolchain('android-ndk-r11c-api-21-mips', 'Unix Makefiles'),
     Toolchain('android-ndk-r11c-api-21-mips64', 'Unix Makefiles'),
+    Toolchain('emscripten-cxx11', 'Unix Makefiles'),
     Toolchain('raspberrypi2-cxx11', 'Unix Makefiles')
 ]
 
--- a/compiler/emscripten.cmake
+++ b/compiler/emscripten.cmake
@@ -1,0 +1,21 @@
+# Copyright (c) 2016, Alexandre Pretyman
+# All rights reserved.
+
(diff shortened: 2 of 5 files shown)


ruslo:
> I'm not sure I understood your comments here

If you will have several versions of SDK you have to do `export EMSCRIPTEN=/path/to/sdk/A` before build. Alternatively you can set in your `.bashrc`:
```
export EMSCRIPTEN_A=/path/to/sdk/A
export EMSCRIPTEN_B=/path/to/sdk/B
```
and load the correct version in toolchain (I probably was wrong about `polly.py`, toolchain should works fine for your case):
```cmake
string(COMPARE EQUAL "$ENV{EMSCRIPTEN_${EMSCRIPTED_SDK_VERSION}}" "" _no_env)
if(_no_env)
  # check ENV{EMSCRIPTEN}, report error if not found
else()
  set(ENV{EMSCRIPTEN} "$ENV{EMSCRIPTEN_${EMSCRIPTED_SDK_VERSION}}")
endif()
```

This is how iOS/OSX toolchains load different version of Xcode, this is how Android toolchains switched between several NDKs, Ninja/NMake toolchains do work a little bit different but with the similar idea - read `VC<visual-studio-version>COMNTOOLS` environment variable.

It's not a problem if you have only one version, just an idea for the future.

PR #121, opened 2017-01-28
Add “—output build_tree” option (i.e., cmake —B/path/to/build/folder)

headupinclouds:
* report error if specified directory does not exist
* report error if specified directory is not writeable
* (else) set operating “current” directory to specified build_tree path and proceed

Note: Use “output” name instead of “build” to avoid confusion with `cmake —build`

### DIFF
--- a/bin/polly.py
+++ b/bin/polly.py
@@ -70,7 +70,7 @@ parser.add_argument(
 
 parser.add_argument(
     '--output',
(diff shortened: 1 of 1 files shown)


headupinclouds:
Github thinks travis is failing.  I've checked the travis logs and don't see any errors.  Thoughts?

PR #122, opened 2017-01-29
add --output option to specify location of build tree

headupinclouds:
* report error if specified directory does not exist
* report error if specified directory is not writeable
* (else) set operating “current” directory to specified build_tree path and proceed
Note: Use “output” name instead of “build” to avoid confusion with cmake —build

### DIFF
--- a/bin/polly.py
+++ b/bin/polly.py
@@ -70,7 +70,7 @@ parser.add_argument(
 
 parser.add_argument(
     '--output',
-    help="Project build directory (directory with CMakeLists.txt)"
+    help="Project build directory (i.e., cmake --B)"
 )
 
 parser.add_argument('--test', action='store_true', help="Run ctest after build")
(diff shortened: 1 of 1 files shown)


headupinclouds:
This time example-01 gcc + Debug needs a kick.  I don't have permission to restart individual jobs.  Please restart this one when you have a chance.

bin/polly.py
```diff
@@ -68,6 +68,11 @@
     help="Project home directory (directory with CMakeLists.txt)"
 )
 
+parser.add_argument(
+    '--output',
+    help="Project build directory (i.e., cmake --B)"
```
ruslo: It's `cmake -B`, not `--B`.

ruslo: https://github.com/ruslo/polly/commit/ca03b1b600bf5718a5d00e95c1921219f404ba8f

headupinclouds: Thanks.

PR #127, opened 2017-02-28
Cache environment variables so that regenerated builds use the same value

crujose:
I normally don't keep these variables defined for my environment.  This makes working with the generated builds a bit of a pain since modifying a cmake file will cause the build to fail when it goes to regenerate.  By caching the environment variable it will simply reuse that value.

### DIFF
--- a/os/android.cmake
+++ b/os/android.cmake
@@ -14,15 +14,18 @@ if(_is_empty)
   polly_fatal_error("ANDROID_NDK_VERSION is not defined")
 endif()
 
-set(_env_ndk "$ENV{ANDROID_NDK_${ANDROID_NDK_VERSION}}")
-string(COMPARE EQUAL "${_env_ndk}" "" _is_empty)
+
+if(DEFINED ENV{ANDROID_NDK_${ANDROID_NDK_VERSION}})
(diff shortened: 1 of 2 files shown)


ruslo:
> I normally don't keep these variables defined for my environment

What is the point of this? You can just keep them in `.bashrc`.

crujose:
Working in large teams across multiple platforms, I've found it's best to keep things simple by removing manual setup whenever possible.  In this case, it's easy to write a wrapper build script that sets the environment variables and generates the build.

ruslo:
> I've found it's best to keep things simple by removing manual setup whenever possible

Why you have to do it manually?

So instead of setting environment variable once you have to add it to every Android build? Make no sense to me. Also such approach will not work with Hunter since `POLLY_ANDROID_NDK` [will not be forwarded][1] to 3rd parties:
```
Toolchain file is the only way to apply global settings for 3rd party projects in Hunter.
Only CMAKE_TOOLCHAIN_FILE will be forwarded for all packages, neither standard
CMAKE_CXX_COMPILER/CMAKE_CXX_FLAGS nor custom variables
like ANDROID_FOO=boo will be applied globally.
```

[1]: https://docs.hunter.sh/en/latest/overview/customization/toolchain-id.html

PR #139, opened 2017-05-03
Make cmake binary configurable

jcftang:
I needed this on centos7 with epel, the cmake version 3 binary is named cmake3 and not cmake.

### DIFF
--- a/bin/polly.py
+++ b/bin/polly.py
@@ -196,6 +196,11 @@ parser.add_argument(
     help='Timeout for CTest'
 )
 
+parser.add_argument(
+    '--cmake',
+    help="CMake binary (cmake or cmake3)"
+)
+
 args = parser.parse_args()
 
 polly_toolchain = detail.toolchain_name.get(args.toolchain)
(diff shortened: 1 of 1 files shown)


bin/polly.py
```diff
@@ -196,6 +196,11 @@ def PositiveInt(string):
     help='Timeout for CTest'
 )
 
+parser.add_argument(
+    '--cmake',
+    help="CMake binary (cmake or cmake3)",
```
ruslo: `,` at the end is redundant

bin/polly.py
```diff
@@ -309,19 +314,24 @@ def PositiveInt(string):
     cdir, args.verbosity, args.discard, args.tail, polly_toolchain
 )
 
+if args.cmake:
+    CMAKE_BIN = args.cmake
```
ruslo: 2 spaces indent

bin/polly.py
```diff
@@ -309,19 +314,24 @@ def PositiveInt(string):
...
+else:
+    CMAKE_BIN = 'cmake'
```
ruslo: please rename to `cmake_bin`, `CMAKE_BIN` assume it something global

bin/polly.py
```diff
@@ -309,19 +314,24 @@ def PositiveInt(string):
...
+
 if os.name == 'nt':
   # Windows
-  detail.call.call(['where', 'cmake'], logging)
+  detail.call.call(['where', CMAKE_BIN], logging)
```
ruslo: If `CMAKE_BIN` is a full path to `cmake` just check that this file exists.

jcftang: I wasn't too sure what you meant by this, so I added the check just before this if block

bin/polly.py
```diff
@@ -309,19 +314,28 @@ def PositiveInt(string):
...
+if args.cmake:
+  cmake_bin = args.cmake
+else:
+  cmake_bin = 'cmake'
+
+if os.path.isabs(cmake_bin):
+  if not os.path.exists(cmake_bin):
+    sys.exit("CMake binary not found: {}".format(cmake_bin))
+
 if os.name == 'nt':
```
ruslo: if `cmake_bin` is absolute path then it make no sense to run `where`/`which`. Actually `where` will not work. Should be like:
```python
if os.path.isabs(cmake_bin):
  # check exists
else:
  if os.name == 'nt':
    # where
  else:
    # which
```

jcftang:
I did a bit more testing in a clean environment and fixed a case where I missed a call to cmake

Issue #62, opened 2016-01-14
Non UTF8 console encoding


ruslo:
When console encoding is not UTF8 (like Windows + non-English locale) then `build.py` script will throw exception UnicodeDecodeError.

Line to blame:
- https://github.com/ruslo/polly/blob/c343d0ad3fe7810c3482c84e2a8b1e8ce8685861/bin/detail/call.py#L17

Originally from:
- https://github.com/ruslo/hunter/issues/298

ruslo:
Workaround: https://github.com/ruslo/hunter/issues/298#issuecomment-171387754

ruslo:
Fixed by https://github.com/ruslo/polly/pull/141

PR #140, opened 2017-05-03
Fix some unicode issues with logging

jcftang:
I was seeing hanging behaviour. When I have this patch my builds succeed.

### DIFF
--- a/bin/detail/call.py
+++ b/bin/detail/call.py
@@ -38,7 +38,7 @@ def tee(infile, discard, log_file, console=None):
       s = s.replace('\t', '  ')
       s = s.rstrip() # strip spaces and EOL
       s += '\n' # append stripped EOL back
-      log_file.write(s)
+      log_file.write(s.encode('utf8'))
       if console is None:
         continue
       if discard is None:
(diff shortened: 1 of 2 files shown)


ruslo:
I've seen some problems with encoding before but it was error from Python, not hanging:
* https://github.com/ruslo/polly/issues/62

I'm not sure I understand all parts, can you please explain in details what is happening? I guess leaving comments with link to documentation would be nice.

jcftang:
My build process generates output which causes this to occur...

```
Exception in thread Thread-6:
Traceback (most recent call last):
  File "/usr/lib64/python3.4/threading.py", line 911, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.4/threading.py", line 859, in run
    self._target(*self._args, **self._kwargs)
  File "/home/build/polly/bin/detail/call.py", line 41, in fanout
    log_file.write(s)
UnicodeEncodeError: 'ascii' codec can't encode character '\xf6' in position 75: ordinal not in range(128)
```

AFAIK the build process for the project that I am working has many different languages which get emitted by whatever compiler/preprocessor that I am using. This output causes a crashing in one of the threads that polly.py spawns and it then causes the script to just wait and never succeed.

ruslo:
Please try fix from https://github.com/ruslo/polly/pull/141

jcftang:
FYI the patch looks good to me, will test later.

PR #141, opened 2017-05-03
Fix encoding

ruslo:
Use 'sys.stdout.encoding' for console and 'utf-8' for log file


jcftang:
FYI I've just tested this and it's addressed the issue that I was seeing.

ruslo:
Please open new issue with info about OS and Python. I guess the string with non-ascii characters is important too. I want to reproduce this bug.

PR #143, opened 2017-05-04
Teach polly.py about different cpack and ctest binaries

jcftang:
This allows me to do...

```
./polly/bin/polly --cmake cmake3 --jobs 8 --pack TBZ2 --reconfig --cpack cpack3
```

Where previously the script expected cpack only as the binary.

### DIFF
--- a/bin/polly.py
+++ b/bin/polly.py
@@ -201,6 +201,16 @@ parser.add_argument(
     help="CMake binary (cmake or cmake3)"
 )
 
+parser.add_argument(
+    '--cpack',
+    help="CPack binary (cpack or cpack3)"
+)
+
+parser.add_argument(
+    '--ctest',
+    help="CTest binary (ctest or ctest3)"
+)
+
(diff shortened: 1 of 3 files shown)


bin/polly.py
```diff
@@ -452,7 +457,17 @@ def PositiveInt(string):
...
+    if args.cpack:
+      cpack_bin = args.cpack
+    else:
+      cpack_bin = 'cpack'
+
+    if os.path.isabs(cpack_bin):
+      if not os.path.exists(cpack_bin):
+        sys.exit("CPack binary not found: {}".format(cmake_bin))
+
+    detail.pack_command.run(args.config, logging, cpack_generator, cpack_bin, cmake_bin)
```
ruslo: I guess `run` should be updated?

jcftang: Apologies on that, I missed that file and was in the middle of teaching the script about ctest_bin as well as I was hitting that case as I use this script.

PR #151, opened 2017-06-08
ios-10-3-dep-8-0-bitcode

ikliashchou:
New toolchain variant with 8.0 deployment target and Bitcode support


ios-10-3-dep-8-0-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
...
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
+
+set(IPHONEOS_ARCHS armv7;armv7s;arm64)
+set(IPHONESIMULATOR_ARCHS i386;x86_64)
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx11.cmake")
+
+set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE YES)
```
ruslo: This flag is not "visible" by toolchain-id calculation module in Hunter (there is no C++ macro to detect this feature). In this case we have to mark toolchain explicitly. Please move this code to separate module `flags/bitcode.cmake` and add `list(APPEND HUNTER_TOOLCHAIN_UNDETECTABLE_ID "bitcode")`.

Example:
* https://github.com/ruslo/polly/blob/8d8b688ea809743a7bca80e34822f9395fbcd211/flags/static.cmake#L15-L17

ios-10-3-dep-8-0-bitcode.cmake
```diff
@@ -0,0 +1,45 @@
...
+
+set(IPHONEOS_ARCHS armv7;armv7s;arm64)
+set(IPHONESIMULATOR_ARCHS i386;x86_64)
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx11.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/bitcode.cmake")
+
+set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE YES)
```
ruslo: Should go to `flags/bitcode.cmake`

ios-10-3-dep-8-0-bitcode.cmake
```diff
@@ -0,0 +1,45 @@
...
+set(CMAKE_MACOSX_BUNDLE YES)
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
+
+set(IPHONEOS_ARCHS armv7;armv7s;arm64)
+set(IPHONESIMULATOR_ARCHS i386;x86_64)
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx11.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/bitcode.cmake")
```
ruslo: Since we have BITCODE=OFF in `os/iphone.cmake` please add a comment:
```cmake
include("${CMAKE_CURRENT_LIST_DIR}/flags/bitcode.cmake") # after os/iphone.cmake
```

PR #160, opened 2017-07-21
Docs: Raspberry Pi

ruslo:


docs/toolchains/raspberry-pi.rst
```diff
@@ -0,0 +1,38 @@
...
+
+  > sudo apt-get install python3
+  > sudo apt-get install g++
+  > sudo apt-get install cmake
+
+Use ``gcc`` toolchain, e.g.:
+
+.. code-block:: none
+
+  > polly.py --toolchain gcc --verbose --config Release
```
ruslo: @pretyman I'm experimenting with Raspberry Pi currently. Looks like `gcc` toolchain just work for me. Why do we need [raspberrypi2-cxx](https://github.com/ruslo/polly/blob/master/raspberrypi2-cxx11.cmake) toolchain? Is it for cross-compiling? I.e. when you compiling code on Linux x86_64 and targeting Raspbian ARM?

pretyman: Yep, cross compiling. All contributions from me were for cross compiling. Compiling on the device has been historically slow, so I always took the cross compilation route. Maybe on the Pi 2 or 3 it is ok, but cross compiling we can also get newer compilers from crosstool-ng.

ruslo: > Yep, cross compiling

@pretyman Can't get it work. I should use this repository with tools: https://github.com/raspberrypi/tools, right? I think there is problem with flags defined here:
* https://github.com/ruslo/polly/blob/f7dd93ab12d0b9e20dcf1146c1c90ce8b836272b/os/raspberry-pi3.cmake#L12
* https://github.com/ruslo/polly/blob/f7dd93ab12d0b9e20dcf1146c1c90ce8b836272b/os/raspberry-pi-hardfloat.cmake#L12

Where they came from?

pretyman: I compile my own toolchain with crosstool-ng, the flags you refer to are specific for the Pi 3's hardware and the toolchain must support them. I believe the toolchain on https://github.com/raspberrypi/tools is generic to work with Pi 2 and 3 (possibly 1 as well) and won't support those flags.

Try using the flags of Raspberry Pi 2 to compile for the 3, it works. Sorry there are no Docs for that :S

pretyman: I have a toolchain for OSX and Pi 2 here: https://github.com/pretyman/raspberrypi2-mac-crosscompiler 	based on GCC 4.9.2

Been using a toolchain for Pi 3 based on GCC 6.3.1 which support those flags, if you want I can put it on github

docs/toolchains/raspberry-pi.rst
```diff
@@ -0,0 +1,71 @@
...
+
+  > git clone https://github.com/raspberrypi/tools raspberrypi-tools
+
+Save paths in environment variables:
+
+.. code-block:: none
+
+  > export RASPBERRYPI_CROSS_COMPILE_TOOLCHAIN_PATH=/.../raspberrypi-tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin
+  > export RASPBERRYPI_CROSS_COMPILE_TOOLCHAIN_PREFIX=arm-linux-gnueabihf
+  > export RASPBERRYPI_CROSS_COMPILE_SYSROOT=/.../raspberrypi-tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/arm-linux-gnueabihf/libc
```
ruslo: >  Can't get it work. I should use this repository with tools: https://github.com/raspberrypi/tools, right? I think there is problem with flags defined here

Problem was with sysroot. This value works fine. I've added `raspberrypi3-cxx11` toolchain.

docs/toolchains/raspberry-pi.rst
```diff
@@ -0,0 +1,98 @@
...
+  > polly.py --toolchain raspberrypi3-cxx11 --verbose --config Release
+
+OSX
+~~~
+
+Download tools:
+
+.. code-block:: none
+
+  > git clone https://github.com/pretyman/raspberrypi2-mac-crosscompiler raspberrypi-tools
```
ruslo: I've tested your tools on my OSX machine, worked fine. I've added instructions to documentation.

> For now it looks pretty hairy and confusing. Need to clarify all details, probably we should use different names for these toolchains.

False alarm, looks solid at this moment.

docs/toolchains/raspberry-pi.rst
pretyman: That's nice to know that the cross compile from `raspberrypi-tools` support the Pi 3 flags, thanks!

docs/toolchains/raspberry-pi.rst
pretyman: > What is the reason?

Just to get access to the latest gcc with C++14 features.

I'm happy to know that everything worked out in the end 👍

PR #161, opened 2017-07-25
add gfortran to clang compiler

NeroBurner:
quickest way I could think of to set the Fortran compiler for clang builds

should fix the `clang-libstdcxx` build over at https://github.com/ruslo/hunter/pull/898#issuecomment-317642751


ruslo:
Closing for https://github.com/ruslo/polly/pull/163

PR #163, opened 2017-07-25
add compiler/gfortran

NeroBurner:
based on PR https://github.com/ruslo/polly/pull/162

adds the compiler/gfortran file and uses it in `clang-libstdcxx`

alternative to https://github.com/ruslo/polly/pull/161

### DIFF
--- a/compiler/gfortran.cmake
+++ b/compiler/gfortran.cmake
@@ -1,0 +1,25 @@
+# Copyright (c) 2017, NeroBurner
+# All rights reserved.
+
+if(DEFINED POLLY_COMPILER_Fortran_CMAKE)
+  return()
+else()
+  set(POLLY_COMPILER_Fortran_CMAKE 1)
+endif()
+
+find_program(CMAKE_Fortran_COMPILER gfortran)
+
+if(NOT CMAKE_Fortran_COMPILER)
+  polly_status_print("gfortran not found")
+else()
+  set(
+      CMAKE_Fortran_COMPILER
+      "${CMAKE_Fortran_COMPILER}"
+      CACHE
+      STRING
+      "Fortran compiler"
+      FORCE
+  )
+endif()
+
+
--- a/clang-libstdcxx.cmake
+++ b/clang-libstdcxx.cmake
@@ -21,3 +21,5 @@ include("${CMAKE_CURRENT_LIST_DIR}/library/std/libstdcxx.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx11.cmake")
 
 include("${CMAKE_CURRENT_LIST_DIR}/os/osx.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/gfortran.cmake")


compiler/gfortran.cmake
```diff
@@ -0,0 +1,23 @@
...
+if(DEFINED POLLY_COMPILER_Fortran_CMAKE)
+  return()
+else()
+  set(POLLY_COMPILER_Fortran_CMAKE 1)
+endif()
+
+find_program(CMAKE_Fortran_COMPILER gfortran)
+
+if(NOT CMAKE_Fortran_COMPILER)
+  polly_fatal_error("gfortran not found")
```
ruslo: This toolchain can be used for projects without Fortran requirement so we should not crash here.

NeroBurner:
rebased and updated to print a status message instead of an error message

NeroBurner:
squashed

NeroBurner:
I can't reproduce this error. I tried it with the following `CMakeLists.txt` file

```
cmake_minimum_required(VERSION 3.0)

function(polly_add_cache_flag var_name flag)
  set(spaced_string " ${${var_name}} ")
  string(FIND "${spaced_string}" " ${flag} " flag_index)
  if(NOT flag_index EQUAL -1)
    return()
  endif()
  string(COMPARE EQUAL "" "${${var_name}}" is_empty)
  if(is_empty)
    # beautify: avoid extra space at the end if var_name is empty
    set("${var_name}" "${flag}" CACHE STRING "" FORCE)
  else()
    set("${var_name}" "${flag} ${${var_name}}" CACHE STRING "" FORCE)
  endif()
endfunction()

# if this line is activated gfortran breaks
#polly_add_cache_flag(CMAKE_Fortran_FLAGS "-stdlib=libstdc++")

polly_add_cache_flag(CMAKE_CXX_FLAGS "-stdlib=libstdc++")
polly_add_cache_flag(CMAKE_EXE_LINKER_FLAGS "-stdlib=libstdc++")
polly_add_cache_flag(CMAKE_SHARED_LINKER_FLAGS "-stdlib=libstdc++")

project(example_project CXX Fortran)
```

This are all the flags that I could find, that `libstdc++` is set. I'm only able to break the `gfortran` compiler if I set `CMAKE_Fortran_FLAGS`, which is not done by the toolchain

ruslo:
> I can't reproduce this error

Try https://github.com/forexample/fortran-simple with `polly.py --toolchain clang-libstdcxx`

NeroBurner:
> Try https://github.com/forexample/fortran-simple with polly.py --toolchain clang-libstdcxx

with your example AND polly I can reproduce the error. Now how do we find what the difference is between my simple example and your simple polly example?

NeroBurner:
calling cmake by hand and supplying the polly toolchain file the same `/usr/bin/gfortran  -- broken` error occurs
```
cmake -H. -B_build -DCMAKE_TOOLCHAIN_FILE=/home/gc/repos/hunter/polly/clang-libstdcxx.cmake
```

ruslo:
I think your code is not equivalent to loading flags with toolchain.
For example toolchain loaded several times in CMake's internals:
* http://cgold.readthedocs.io/en/latest/tutorials/project.html#tools-discovering

Just for your info without `polly.py` issue reproduced by:
```
> rm -rf _builds
> cmake -H. -B_builds -DCMAKE_TOOLCHAIN_FILE=/.../polly/clang-libstdcxx.cmake
```

NeroBurner:
I recreated a toolchain file and gfortran breaks by adding the following line

```
polly_add_cache_flag(CMAKE_EXE_LINKER_FLAGS "-stdlib=libstdc++")
```

minimal `toolchain-clang-gfortran.cmake` File
```
set(CMAKE_C_COMPILER clang)
set(CMAKE_CXX_COMPILER clang++)
set(CMAKE_Fortran_COMPILER gfortran)

set(CMAKE_EXE_LINKER_FLAGS "-stdlib=libstdc++" CACHE STRING "" FORCE)

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

ruslo:
> breaks by adding the following line

That's the whole point of this toolchain :) Otherwise it's the same as `libcxx` (Clang + `libc++`), which is not working on Linux even for C++ projects.

PR #180, opened 2017-09-25
add toolchain for mac os 10.10 using default sdk

arrtchiu:
Copied and adapted the contents of `os/osx.cmake` in the bottom of the file to avoid using the specifically-versioned Mac OS SDK. Not sure how that fits with how you'd like to structure this, so we may want to adjust before merging.


ruslo:
Isn't it the same as `osx-10-10-dep-10-7`:
* https://github.com/ruslo/polly/blob/master/osx-10-10-dep-10-7.cmake
?

arrtchiu:
Let's give this a miss - it's better to specify an exact SDK rather than "system default"

PR #181, opened 2017-09-25
Ios11

pthom:
Hi 

This is a pull request that tries to add support for iOS 11 (since iOS 11 removes support for 32 bits app). I do not know if you were already working on it on your side.
It was succesfully tested on my with a Mac that has the newest XCode 9.0 installed, andl only iPhone11 sdk.

PS : I am sorry if this PR mixes cxx14 and iOS11 

Summary : added `flags/cxx14.make`, added `ios-11-0.cmake` and `ios-nocodesign-11-0.cmake` 

Main differences : 
* use flags/cxx14.cmake 
* removed armv7 and armv7s support since 32 bits is not supported anymore by iOS 11 
* also removed simulator arch i386 in favor of the 64 bits simulator arch x86_64

Notes : 
* I left`ios-nocodesign.cmake` unchanged, but it still refers to an older sdk (iOS8.1)
* I wondered if it would be a good idea to add a toolchain ios-latest.cmake that would automatically search for the highest SDK inside /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/
What do you think ?

### DIFF
--- a/flags/cxx14.cmake
+++ b/flags/cxx14.cmake
@@ -1,0 +1,19 @@
+# Copyright (c) 2013, Ruslan Baratov
+# All rights reserved.
+
+if(DEFINED POLLY_FLAGS_CXX14_CMAKE)
+return()
+else()
+set(POLLY_FLAGS_CXX14_CMAKE 1)
+endif()
+
+include(polly_add_cache_flag)
+
+string(COMPARE EQUAL "${ANDROID_NDK_VERSION}" "" _not_android)
+
+# TODO: test other platfroms, CMAKE_CXX_FLAGS_INIT should work for all
+if(_not_android)
+polly_add_cache_flag(CMAKE_CXX_FLAGS "-std=c++14")
+else()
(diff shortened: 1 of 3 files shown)


ruslo:
Thanks. I will download Xcode 9 and test it.

pthom:
You might have some issues because of a bug in llvm :
https://forums.developer.apple.com/thread/87814
I had to change some of our cmakelists in order to remove `-isystem` on `sysroot/usr/include`

Hope this will be solved soon

ruslo:
> I wondered if it would be a good idea to add a toolchain ios-latest.cmake that would automatically search for the highest SDK inside

Not highest but default. User can switch default Xcode to highest if needed. Toolchain name can be `ios.cmake` and `ios-nocodesign.cmake`. Pull request welcome.

> I leftios-nocodesign.cmake unchanged, but it still refers to an older sdk (iOS8.1)

Yes, I guess it should be named `ios-nocodesign-8-1`.

PR #182, opened 2017-09-25
 POLLY_IOS_DEVELOPMENT_TEAM can also be a cmake variable

pthom:
POLLY_IOS_BUNDLE_IDENTIFIER and POLLY_IOS_DEVELOPMENT_TEAM can be set
 either through an environment variable or through a regular cmake variable

 This enables to create a wrapper toolchain that includes these variables i.e for example:

````
set(POLLY_IOS_DEVELOPMENT_TEAM "XXXXXX")
set(POLLY_IOS_BUNDLE_IDENTIFIER "com.example.pollyenabler")
include("${CMAKE_CURRENT_LIST_DIR}/polly-toolchains/ios-11-0.cmake")
`````

### DIFF
--- a/utilities/polly_ios_bundle_identifier.cmake
+++ b/utilities/polly_ios_bundle_identifier.cmake
@@ -9,11 +9,17 @@ endif()
 
 include(polly_status_debug)
 
-string(COMPARE EQUAL "$ENV{POLLY_IOS_BUNDLE_IDENTIFIER}" "" _is_empty)  
-if(_is_empty)
-  set(MACOSX_BUNDLE_GUI_IDENTIFIER "com.example") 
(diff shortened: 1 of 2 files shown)


ruslo:
`MACOSX_BUNDLE_GUI_IDENTIFIER` set for `try_compile` internal tests, you can set this variable after `project` command in your `CMakeLists.txt`. Using `POLLY_IOS_DEVELOPMENT_TEAM` makes no sense in my opinion. It's a global trait, why do you need it per local project? Also it doesn't work well with Hunter because such variable will not be forwarded to 3rd party builds:
* https://docs.hunter.sh/en/latest/overview/customization/toolchain-id.html

pthom:
Ok, you are right :-)
 I will close this PR

PR #184, opened 2017-09-26
add sdk-10-12 targeting osx-10-10

arrtchiu:


bin/detail/toolchain_table.py
```diff
@@ -395,7 +395,8 @@ def verify(self):
       Toolchain('osx-10-12', 'Xcode', osx_version='10.12'),
       Toolchain('osx-10-12-lto', 'Xcode', osx_version='10.12'),
       Toolchain('osx-10-12-cxx98', 'Xcode', osx_version='10.12'),
-      Toolchain('osx-10-10-dep-10-7', 'Xcode', osx_version='10.10'),
+      Toolchain('osx-10-10-dep-10-7', 'Xcode', osx_version='10.7'),
```
arrtchiu: I think `osx_version=` here was not quite right before. Could you please confirm?

ruslo: `osx_version` is only used to read `OSX_X_Y_DEVELOPER_DIR`:
* https://github.com/ruslo/polly/blob/4449eaa23dbadace7fee6b6f78ec74e078072dc3/bin/detail/osx_dev_root.py

`osx_version=10.7` will pick Xcode 4.6.3 which don't have SDK 10.10. By using `osx_version=10.10` we will pick Xcode 6 which have SDK 10.10 and will add `-mmacosx-version-min=10.7` flag to compiler.

Table with default SDK is here:
* http://polly.readthedocs.io/en/latest/toolchains/ios.html

bin/detail/toolchain_table.py
```diff
@@ -395,7 +395,8 @@ def verify(self):
...
+      Toolchain('osx-10-12-dep-10-10', 'Xcode', osx_version='10.10'),
```
ruslo: Use `osx_version=10.12`.

PR #187, opened 2017-09-27
ios11-sdk with deployment-target ios9

arrtchiu:


ios-11-0-dep-9-0-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
...
+
+include(polly_fatal_error)
+
+# Fix try_compile
+include(polly_ios_bundle_identifier)
+
+set(CMAKE_MACOSX_BUNDLE YES)
+set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
+
+set(IPHONEOS_ARCHS armv7;armv7s;arm64)
```
ruslo: 32 bit architectures dropped since iOS 11.0:
* https://github.com/ruslo/polly/blob/2f1b497df080164f459a7733069267e7eafff03c/ios-11-0.cmake#L33-L34

ios-11-0-dep-9-0-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
...
+set(IPHONESIMULATOR_ARCHS i386;x86_64)
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx11.cmake")
```
ruslo: I've already accepted iOS 11.0 with C++14 as default:
* https://github.com/ruslo/polly/pull/181/files

To keep things consistent please use C++14 or rename toolchain to `ios-11-0-dep-9-0-bitcode-cxx11`.

ios-11-0-dep-9-0-bitcode.cmake
arrtchiu: I think it's still valid to build 32-bit, as long as we're targeting <= iOS10 (but can still build with newer SDK)

> 32-bit Apps
> Deprecations
> 32-bit apps no longer run on iOS 11.
> To include a 32-bit slice in your app, target iOS 10 or earlier.

https://developer.apple.com/library/content/releasenotes/General/RN-iOSSDK-11/index.html

ruslo:
> I think it's still valid to build 32-bit, as long as we're targeting <= iOS10

Okay, I see.

Looks good but I can't merge because of the conflicts.

PR #188, opened 2017-09-27
Rename openwrt to ddwrt

arrtchiu:
toolchains added as "openwrt" are actually dd-wrt :)


arrtchiu:
I messed this up. Will do another PR shortly.

PR #196, opened 2017-10-28
add —cache option

headupinclouds:
* CMake -C <initial-cache> = Pre-load a script to populate the cache.
* minor typo/grammar fix


ruslo:
Thanks. Note that if `CMakeCache.txt` exists and `--fwd` arguments haven't changed then generate step will not be triggered. It means for example that if you change content of `--cache` file you have to do `--reconfig`.

PR #197, opened 2017-10-28
add —config-all option

headupinclouds:
Add shorthand for: --config <type> --fwd HUNTER_CONFIGURATION_TYPES=<type>:

i.e., —config-all Release ==

  `-DCMAKE_BUILD_TYPE=Release`
  `-DHUNTER_CONFIGURATION_TYPES=Release`

### DIFF
--- a/bin/polly.py
+++ b/bin/polly.py
@@ -78,6 +78,11 @@ parser.add_argument(
     help="Project build directory (i.e., cmake -B)"
 )
 
+parser.add_argument(
+    '--cache',
+    help="CMake -C <initial-cache> = Pre-load a script to populate the cache."
+)
+
 parser.add_argument('--test', action='store_true', help="Run ctest after build")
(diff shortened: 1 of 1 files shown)


bin/polly.py
```diff
@@ -350,7 +361,7 @@ def PositiveInt(string):
     build_dir_option
 ]
 
-if args.config and not toolchain_entry.multiconfig:
+if (args.config and not toolchain_entry.multiconfig) or args.config_all:
```
headupinclouds: @ruslo : I left the existing logic for --config as it was.  By default, if it is a multiconfig toolchain it does not set `-DCMAKE_BUILD_TYPE=<type>`.  In the case of `--config-all` it will set `-DCMAKE_BUILD_TYPE=<type>.

headupinclouds:
This shorthand will help simplify hunter+polly readme files.

PR #210, opened 2017-12-04
ios-nocodesign-11-1-dep-9-0-bitcode toolchain added

stefan-sedlak:


ios-nocodesign-11-1-dep-9-0-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
...
+include(polly_fatal_error)
+
+# Fix try_compile
+include(polly_ios_bundle_identifier)
+
+set(CMAKE_MACOSX_BUNDLE YES)
+
+include("${CMAKE_CURRENT_LIST_DIR}/flags/ios_nocodesign.cmake")
+
+set(IPHONEOS_ARCHS armv7;arm64)
```
ruslo: If you don't want to have `armv7s` architecture you should rename this toolchain. See same iOS 10.3 toolchain:
* https://github.com/ruslo/polly/blob/b4bbb3a12be67736a5529c53a9615e295a719cc0/ios-10-3-dep-9-0-bitcode.cmake#L36-L37

stefan-sedlak: As you wish, armv7s was dropped from Xcode as developers had no profit from such arch. It increased my fat-lib by 20% from 40MB to 50MB. What shall be proper name? list of 4 architectures? or do you prefer adding armv7s

ruslo: > armv7s was dropped from Xcode as developers had no profit from such arch

It's really not about been dropped or useful. If iOS 11.1 will not have it and the same toolchains only with iOS 10.3 will have it then everything will work/look inconsistently.

> What shall be proper name?

You can add `wo-armv7s`, like this one:
* https://github.com/ruslo/polly/blob/master/ios-10-1-wo-armv7s.cmake

> do you prefer adding armv7s

It's up to you.

stefan-sedlak: Thx, I will use wo-armv7s to keep consistency. I found ugly thing. In compile flags I found both flags -std=c++14 --std=c++11 . I will change as well cxx14 to cxx11 to avoid issues.

stefan-sedlak: Does not it depend it on deployment target 9.0 here? just asking.

ruslo: Similar toolchains signal C++11 explicitly by adding suffix `-cxx11` to the name:
* https://github.com/ruslo/polly/blob/master/ios-11-0-dep-9-0-bitcode-cxx11.cmake

In general there are no hard rules here, just common sense.

ios-nocodesign-11-1-dep-9-0-wo-armv7s-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2017, Ruslan Baratov
+# All rights reserved.
+
+if(DEFINED POLLY_IOS_NOCODESIGN_11_0_DEP_9_0_BITCODE_CMAKE_)
```
ruslo: `_11_1_`

ios-nocodesign-11-1-dep-9-0-wo-armv7s-bitcode.cmake
```diff
@@ -0,0 +1,44 @@
...
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_clear_environment_variables.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+set(IOS_SDK_VERSION 11.1)
+set(IOS_DEPLOYMENT_SDK_VERSION 9.0)
+
+set(POLLY_XCODE_COMPILER "clang")
+polly_init(
+    "iOS ${IOS_SDK_VERSION} / Deployment ${IOS_DEPLOYMENT_SDK_VERSION} / Universal (iphoneos + iphonesimulator) / \
```
ruslo: Add list of libraries here.

https://github.com/ruslo/polly/blob/6b4b3d2b866fee571d900f9f5dfcb4fa4eab863f/ios-10-0-wo-armv7s.cmake#L19

stefan-sedlak:
I close PR, I will provide new PR with less commits. See you soon :)

ruslo:
> I will provide new PR with less commits

You can squash and to `git push -f` (force). No need to close pull request.

PR #214, opened 2017-12-14
MSVC C++17

dvirtz:
Added toolchains for VS 15 2017 with C++17 switch

### DIFF
--- a/flags/cxx17.cmake
+++ b/flags/cxx17.cmake
@@ -1,0 +1,22 @@
+# Copyright (c) 2013, Ruslan Baratov
+# All rights reserved.
+
+if(DEFINED POLLY_FLAGS_CXX17_CMAKE)
+  return()
+else()
+  set(POLLY_FLAGS_CXX17_CMAKE 1)
+endif()
+
+include(polly_add_cache_flag)
+
+string(COMPARE EQUAL "${ANDROID_NDK_VERSION}" "" _not_android)
+
+
+# TODO: test other platfroms, CMAKE_CXX_FLAGS_INIT should work for all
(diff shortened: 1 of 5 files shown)


flags/cxx17.cmake
```diff
@@ -0,0 +1,22 @@
...
+endif()
+
+include(polly_add_cache_flag)
+
+string(COMPARE EQUAL "${ANDROID_NDK_VERSION}" "" _not_android)
+
+
+# TODO: test other platfroms, CMAKE_CXX_FLAGS_INIT should work for all
+if(HUNTER_CMAKE_GENERATOR MATCHES "^Visual Studio.*$")
+  polly_add_cache_flag(CMAKE_CXX_FLAGS "/std:c++17")
```
ruslo: I think this trick is not working with Visual Studio generator. That's why all flags listed explicitly in `vs-mt` for example:
* https://github.com/ruslo/polly/blob/b9819d2e7f2be813b97654f6cbba20831751fa58/flags/vs-mt.cmake#L11-L14

vs-15-2017-cxx17.cmake
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2015-2017, Ruslan Baratov
```
ruslo: Please add it to toolchain table:
* https://github.com/ruslo/polly/blob/b9819d2e7f2be813b97654f6cbba20831751fa58/bin/detail/toolchain_table.py#L180-L185

dvirtz:
Using `CMAKE_CXX_FLAGS_INIT` instead of `CMAKE_CXX_FLAGS` seems to fix it.

dvirtz:
Thanks

PR #215, opened 2017-12-15
Add Android toolchains (NDK r15c: arm64-v8a, NDK r16b: arm64-v8a, armeabi-v7a, x86)

rsnitsch:
Filenames should be self-explanatory.


rsnitsch:
Okay. I have added the toolchains to the toolchain_table.py.

PR #216, opened 2017-12-17
add gcc-7

headupinclouds:


gcc-7.cmake
```diff
@@ -0,0 +1,21 @@
...
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+polly_init(
+    "gcc 7 / c++11 support"
+    "Unix Makefiles"
+)
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_common.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/gcc-5.cmake")
```
ruslo: `s,gcc-5,gcc-7,g'

headupinclouds: sorry about that one.  good catch.

PR #218, opened 2017-12-22
gcc-aarch64 and gcc suffix

jglanz:
Added linux-gcc-aarch64.cmake 
Added suffix support for GCC layouts (I needed it for -6)

### DIFF
--- a/compiler/gcc-cross-compile.cmake
+++ b/compiler/gcc-cross-compile.cmake
@@ -40,6 +40,17 @@ if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PREFIX not set.")
 endif()
 
+# string(COMPARE EQUAL
+#     "${CROSS_COMPILE_TOOLCHAIN_SUFFIX}"
+#     ""
+#     _is_empty
+# )
+# if(_is_empty)
+#   message("No suffix for toolchain")
+#   set(CROSS_COMPILE_TOOLCHAIN_SUFFIX "")
+# endif()
+
+
 string(COMPARE EQUAL
     "${CROSS_COMPILE_SYSROOT}"
     ""
@@ -48,6 +59,8 @@ string(COMPARE EQUAL
 if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_SYSROOT not set.")
 endif()
+
+
 
 if(POLLY_SKIP_SYSROOT)
   # Do not modify CMAKE_{C,CXX}_FLAGS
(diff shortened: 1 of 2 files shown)


compiler/gcc-cross-compile.cmake
```diff
@@ -40,6 +40,7 @@ if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_TOOLCHAIN_PREFIX not set.")
 endif()
 
+
```
ruslo: Cleanup

compiler/gcc-cross-compile.cmake
```diff
@@ -49,6 +50,8 @@ if(_is_empty)
   polly_fatal_error("CROSS_COMPILE_SYSROOT not set.")
 endif()
 
+
+
```
ruslo: Cleanup

linux-gcc-aarch64.cmake
```diff
@@ -0,0 +1,46 @@
...
+else()
+  set(POLLY_LINUX_GCC_AARCH64_ 1)
+endif()
+
+
+# set system name, this sets the variable CMAKE_CROSSCOMPILING
+set(CMAKE_SYSTEM_NAME Linux)
+
+set(CROSS_COMPILE_TOOLCHAIN_PREFIX aarch64-linux-gnu)
+set(CROSS_COMPILE_TOOLCHAIN_SUFFIX -6)
```
ruslo: Probably should go to the name of toolchain, like `linux-gcc-6-aarch64`.

linux-gcc-aarch64.cmake
```diff
@@ -0,0 +1,46 @@
...
+  set(SYSROOT $ENV{SYSROOT})
+endif()
+
+set(CROSS_COMPILE_SYSROOT ${SYSROOT})
+
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+polly_init(
+    "Linux / gcc / aarch64 / c++11 support"
```
ruslo: `c++17 support` ?

linux-gcc-aarch64.cmake
```diff
@@ -0,0 +1,46 @@
...
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_common.cmake")
+
+
+set(CMAKE_CROSSCOMPILING_EMULATOR qemu-aarch64) # used for try_run calls
+
+include(
+    "${CMAKE_CURRENT_LIST_DIR}/compiler/gcc-cross-compile.cmake"
+)
+
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx17.cmake")
```
ruslo: Currently similar toolchains set C++11 by default so if you want C++17 I think we should name toolchain differently, like `linux-gcc-aarch64-cxx17`.

linux-gcc-aarch64.cmake
```diff
@@ -0,0 +1,46 @@
...
+  set(POLLY_LINUX_GCC_AARCH64_ 1)
+endif()
+
+
+# set system name, this sets the variable CMAKE_CROSSCOMPILING
+set(CMAKE_SYSTEM_NAME Linux)
+
+set(CROSS_COMPILE_TOOLCHAIN_PREFIX aarch64-linux-gnu)
+set(CROSS_COMPILE_TOOLCHAIN_SUFFIX -6)
+set(CROSS_COMPILE_TOOLCHAIN_PATH /usr/bin)
```
ruslo: Currently `CROSS_COMPILE_TOOLCHAIN_PATH` is used in `gcc-musl`/`arm-openwrt-linux-muslgnueabi` and set basing on environment variable that is tied to the toolchain name:
* https://github.com/ruslo/polly/blob/3b5ab745342adbff60ff6f9308bf55aabd19585b/gcc-musl.cmake#L19
* https://github.com/ruslo/polly/blob/3b5ab745342adbff60ff6f9308bf55aabd19585b/arm-openwrt-linux-muslgnueabi.cmake#L37

I think we should follow same approach here:
* Read environment variable `LINUX_GCC_AARCH64_ROOT`
* If not set use default path `/usr/bin` (do we need this?)
* Check path is correct, e.g. `$ENV{GCC_AARCH64_ROOT}/aarch64-linux-gnu-gcc-6` file exist
* If file not found emit error message with clear description of what is expected

compiler/gcc-cross-compile.cmake
ruslo: Cleanup

compiler/gcc-cross-compile.cmake
ruslo: Cleanup

ruslo:

Nadav-Eyesight:
I don't see any aarch64 toolchains.

ruslo:
@Nadav-Eyesight This pull request was closed without merge because of inactivity. Feel free to continue working on.

PR #224, opened 2018-01-25
New Toolchain: Gcc-c14

ckaminski:
Added support for the GCC C++14 toolchain


gcc-c14.cmake
```diff
@@ -0,0 +1,24 @@
...
+
+polly_init(
+    "gcc / c++14 support / C11"
+    "Unix Makefiles"
+)
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_common.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/gcc.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx14.cmake")
```
ruslo: Name of the toolchain doesn't reflect this, probably it's better should be `gcc-cxx14-c11`. `cxx11` part is missing in `gcc-c11` because it's kind of a default for now.

ckaminski: Updated with requested change.

ruslo:
Thanks.

PR #225, opened 2018-02-01
Cxx14 example

ckaminski:
Added gcc-5 and clang-5 C++14 support

Added c++14 testcase.
Sourced from: 
https://stackoverflow.com/a/30940639/51789 
http://en.cppreference.com/w/cpp/utility/tuple/make_tuple

Apologies for the superfluous/undescriptive commits - was attempting to figure out nuances of TravisCI and appveyor.


ckaminski:
My attempts to fix clang seem to not work for OSx, so that test is failing.  I'll see if I can figure out a better way.

examples/04-cxx14/main.cpp
```diff
@@ -0,0 +1,31 @@
+/* 
+ * Source C++ test from: https://stackoverflow.com/a/30940639/51789 
+ */ 
+#include <iostream>
```
ruslo: There is no need to create separate example, you can just add this code to old example with condition `#if __cplusplus >= 201402L`.

PR #226, opened 2018-02-03
add generic clang + libcxx toolchain with -std=c++14

headupinclouds:
This is similar to the libcxx.cmake toolchain but with a c++14 language flag.  It works with the default OSX Apple clang, for example, which has no versioned naming convention used by the llvm builds (i.e., clang-5).


headupinclouds:
NOTE: This should provide a useful default clang toolchain for c++14 hunter packages.

PR #229, opened 2018-02-06
add full android-nd-r16b downloads

headupinclouds:


headupinclouds:
Are the full ndk releases workable as an unoptimized fallback path for CI testing?

PR #240, opened 2018-03-24
android ndk r16b gcc + libc++ toolchain for LTO for MinSizeRel builds

headupinclouds:
NOTE: ndk 17-beta1 still doesn’t support LTO + MinSizeRel builds

It reports:
> "Optimization level must be between 0 and 3"

gcc + libcxx seems to be workable, but this combination can define an “unwind” symbol twice, which can be ignored with the following in the top level project target:

```
set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--allow-multiple-definition")
```

### DIFF
--- a/android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
+++ b/android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
@@ -1,0 +1,43 @@
+# Copyright (c) 2015-2018, Ruslan Baratov
+# Copyright (c) 2015-2018, David Hirvonen
+# Copyright (c) 2015-2018, Alexandre Pretyman
+# All rights reserved.
+
+if(DEFINED POLLY_ANDROID_NDK_R16B_API_19_ARMEABI_V7A_NEON_HID_SECTIONS_LTO_CMAKE_)
+  return()
+else()
+  set(POLLY_ANDROID_NDK_R16B_API_19_ARMEABI_V7A_NEON_HID_SECTIONS_LTO_CMAKE_ 1)
+endif()
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_clear_environment_variables.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+set(ANDROID_NDK_VERSION "r16b")
+set(CMAKE_SYSTEM_VERSION "19")
+set(CMAKE_ANDROID_ARCH_ABI "armeabi-v7a")
+set(CMAKE_ANDROID_ARM_NEON TRUE)
+set(CMAKE_ANDROID_ARM_MODE TRUE) # 32-bit ARM
+set(CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION "4.9") # <major>.<minor>: GCC of specified version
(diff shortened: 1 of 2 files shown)


android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
```diff
@@ -0,0 +1,43 @@
...
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_clear_environment_variables.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+set(ANDROID_NDK_VERSION "r16b")
+set(CMAKE_SYSTEM_VERSION "19")
+set(CMAKE_ANDROID_ARCH_ABI "armeabi-v7a")
+set(CMAKE_ANDROID_ARM_NEON TRUE)
+set(CMAKE_ANDROID_ARM_MODE TRUE) # 32-bit ARM
+set(CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION "4.9") # <major>.<minor>: GCC of specified version
```
ruslo: There is no such variables in similar toolchain:
* https://github.com/ruslo/polly/blob/master/android-ndk-r10e-api-19-armeabi-v7a-neon-hid-sections.cmake

Rename to `*-gcc49-*`?

android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
```diff
@@ -0,0 +1,43 @@
...
+set(CMAKE_ANDROID_STL_TYPE "c++_static") # LLVM libc++ static
```
ruslo: Same here, `libcxx` needed in name. Actually are they compatible?

android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
headupinclouds: Previously gcc was the default.  In r16b clang is the default.  We can still specify a gcc toolchain using this notation: https://cmake.org/cmake/help/v3.7/variable/CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION.html

headupinclouds: > Rename to *-gcc49-*

I'm with you now.  That makes sense.  I'll rename and resubmit.

android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
headupinclouds: It seems so.  I wasn't sure about this until testing it actually.

android-ndk-r16b-api-19-armeabi-v7a-neon-hid-sections-lto.cmake
headupinclouds: Minor correction: use `*-gcc-49-*` to match other toolchains, like: `android-ndk-r11c-api-21-arm64-v8a-gcc-49`

PR #241, opened 2018-03-24
android-ndk-r16b-api-19-gcc-49-armeabi-v7a-neon-hid-sections-lto

headupinclouds:
Follow up to #240 

android ndk r16b gcc + libc++ toolchain for LTO for MinSizeRel builds

NOTE: ndk 17-beta1 still doesn’t support LTO + MinSizeRel builds

It reports:
> "Optimization level must be between 0 and 3"

gcc + libcxx seems to be workable, but this combination can define an “unwind” symbol twice, which can be ignored with the following in the top level project target:

```
set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--allow-multiple-definition")
```


android-ndk-r16b-api-19-gcc-49-armeabi-v7a-neon-hid-sections-lto.cmake
```diff
@@ -0,0 +1,43 @@
...
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+set(ANDROID_NDK_VERSION "r16b")
+set(CMAKE_SYSTEM_VERSION "19")
+set(CMAKE_ANDROID_ARCH_ABI "armeabi-v7a")
+set(CMAKE_ANDROID_ARM_NEON TRUE)
+set(CMAKE_ANDROID_ARM_MODE TRUE) # 32-bit ARM
+set(CMAKE_ANDROID_NDK_TOOLCHAIN_VERSION "4.9") # <major>.<minor>: GCC of specified version
+set(CMAKE_ANDROID_STL_TYPE "c++_static") # LLVM libc++ static
```
ruslo: I guess we need `libcxx` in toolchain name then.

headupinclouds: 👀 👍 Fixed.

PR #247, opened 2018-05-06
Added gcc-6-32bit.cmake

Nadav-Eyesight:
Just a toolchain I found myself needing. I don't know if there's anything else I need to add for the automation, the contributing file doesn't mention anything.


gcc-6-32bit.cmake
```diff
@@ -0,0 +1,21 @@
...
+
+polly_init(
+    "gcc 6 / c++14 support / 32 bit"
+    "Unix Makefiles"
+)
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_common.cmake")
+
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/gcc-6.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx14.cmake")
```
ruslo: gcc toolchains by default comes with C++11 (e.g. `gcc-7`), if you want to have C++14 please rename it to `gcc-6-32bit-cxx14`.

Nadav-Eyesight:
Done. Editing the toolchain table should be a requirement in https://github.com/ruslo/polly/blob/master/CONTRIBUTING.md

compiler/gcc-6-32bit-cxx14.cmake
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2013-2017, Ruslan Baratov
```
ruslo: Please change the name of the file to `compiler/gcc-6.cmake`

Nadav-Eyesight:
I totally forgot to change the line `include("${CMAKE_CURRENT_LIST_DIR}/compiler/gcc-6.cmake")`, which sort of fixed itself now :smile:

PR #259, opened 2018-06-20
Added explicit toolset option for toolchains

Honeybunch:
This allows use of the new toolchain I added which targets VS2017 with the LLVM-vs2014 toolset. This is useful for those of us who like to use Visual Studio and compile with Clang. 

It's also possible that this option could lead to the removal of the "xp" field as another option would be to just explicitly pass the XP toolset for that toolchain. I didn't think I should make that call though.


bin/detail/toolchain_table.py
```diff
@@ -315,6 +317,13 @@ def verify(self):
           arch='amd64',
           vs_version='15'
       ),
+      Toolchain(
+          'vs-15-2017-win64-llvm',
+          'Visual Studio 15 2017 Win64',
+          toolset='LLVM-vs2014',
```
ruslo: I haven't used this yet, don't know the details. So can we do `LLVM-vs2017` too? In this case toolchain name should be `vs-15-2017-win64-llvm-2014` and `vs-15-2017-win64-llvm-2017`.

Honeybunch: Clang 6.0 (latest full release) only provides a toolset named LLVM-vs2014. I can name the toolchain accordingly but there is no LLVM-vs2017

PR #262, opened 2018-06-22
Fix 32 bits support in ios.cmake

lucatorella:
32 bits support was dropped starting from iOS 11.0, but it is still possible, with the iOS 11.x SDK, to build for 32 bits architectures and actually you should do that if the deployment target is < iOS 11.

In other words, it is possible to use the iOS 11.x SDK and target iOS < 11: if that is the case, it is most likely necessary to support 32 bits.

### DIFF
--- a/ios.cmake
+++ b/ios.cmake
@@ -29,19 +29,19 @@ set(CMAKE_MACOSX_BUNDLE YES)
 
 set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
 
-# 32 bits support was dropped from iPhoneSdk11.0
-if(IOS_SDK_VERSION VERSION_LESS "11.0")
-  set(IPHONEOS_ARCHS armv7;armv7s;arm64)
-  set(IPHONESIMULATOR_ARCHS i386;x86_64)
-else()
-  polly_status_debug("iPhone11.0+ SDK detected, forcing 64 bits builds.")
-  set(IPHONEOS_ARCHS arm64)
-  set(IPHONESIMULATOR_ARCHS x86_64)
-endif()
-
 include("${CMAKE_CURRENT_LIST_DIR}/compiler/xcode.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/flags/cxx14.cmake")
+
+# 32 bits support was dropped from iOS 11.0
+if(IOS_DEPLOYMENT_SDK_VERSION VERSION_LESS "11.0")
+  set(IPHONEOS_ARCHS armv7;armv7s;arm64)
+  set(IPHONESIMULATOR_ARCHS i386;x86_64)
+else()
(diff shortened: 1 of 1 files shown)


ios.cmake
```diff
@@ -29,12 +29,12 @@ set(CMAKE_MACOSX_BUNDLE YES)
 
 set(CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "iPhone Developer")
 
-# 32 bits support was dropped from iPhoneSdk11.0
-if(IOS_SDK_VERSION VERSION_LESS "11.0")
+# 32 bits support was dropped from iOS 11.0
+if(IOS_DEPLOYMENT_SDK_VERSION VERSION_LESS "11.0")
```
ruslo: `IOS_DEPLOYMENT_SDK_VERSION` is empty here.

lucatorella: You are right. I didn't realize because in my code I'm doing

```
# set the deployment target
set(IOS_DEPLOYMENT_SDK_VERSION "9.0")

# now include the actual toolchain
include(${CMAKE_CURRENT_LIST_DIR}/polly/ios.cmake)
```

ruslo: I don't see any difference, `IOS_DEPLOYMENT_SDK_VERSION` is still empty.

As a general note name `ios` mean default SDK version and default deployment version hence it's assumed that they are equal. If you want to have latest SDK version and some deployment version I guess toolchain should be named something like `ios-dep-9-0`.

lucatorella: Are you sure? According to what I observed, it's not.`include("${CMAKE_CURRENT_LIST_DIR}/os/iphone.cmake")` set the variable to the default SDK version if it's empty. If it's not empty, it will keep the defined value.

So in the general case it will use the SDK version, but if you define `IOS_DEPLOYMENT_SDK_VERSION` it will use that value.

Answering to your second remark, yes I want to use the default (latest SDK) and 9.0 as deployment target. I think there are two ways to achieve that:
1. create a new `ios-dep-9-0.cmake` basically duplicating `ios.cmake` (what you suggested).
2. use the current `ios.cmake` and set the `IOS_DEPLOYMENT_SDK_VERSION` to 9.0.

The problem with the second approach is that it doesn't work without my patch to `ios.cmake`.

In any case, the current code (or at least the comment) is wrong:
```
# 32 bits support was dropped from iPhoneSdk11.0
if(IOS_SDK_VERSION VERSION_LESS "11.0")
  set(IPHONEOS_ARCHS armv7;armv7s;arm64)
  set(IPHONESIMULATOR_ARCHS i386;x86_64)
else()
  polly_status_debug("iPhone11.0+ SDK detected, forcing 64 bits builds.")
  set(IPHONEOS_ARCHS arm64)
  set(IPHONESIMULATOR_ARCHS x86_64)
endif()
```

The fact that the iOS SDK is 11.0 or greater doesn't imply that 32 bits can be dropped. It depends on the deployment target. 

I understand that in this toolchain the deployment target is, by default, the same as the SDK, but this can be changed (as I did) by setting it.

ruslo:

PR #267, opened 2018-07-25
add cxx17.cmake to match existing cxx11.cmake

madmongo1:


cxx17.cmake
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2013-2017, Ruslan Baratov
+# Copyright (c) 2018, Richard Hodges
+
+# All rights reserved.
+
+if(DEFINED POLLY_CXX17_CMAKE_)
```
ruslo: Please add it to toolchain table.

PR #270, opened 2018-08-06
Add ios-11-4-dep-8-0-arm64-hid-sections-lto-cxx11

headupinclouds:
Based on ios-11-4-dep-9-4-arm64-hid-sections-lto-cxx11


headupinclouds:
Thanks!

PR #294, opened 2018-11-24
add "-fPIC" to CMAKE_Fortran_FLAGS

headupinclouds:
Apply `-fPIC` to `CMAKE_Fortran_FLAGS` as well as `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` so compiled binaries are compatible at link time.

PR posted for initial discussion.

It seems reasonable that flags specified in `polly/flags/*.cmake` should apply to both C/C++ and Fortran, although historically all of these polly flags have really been C/C++/Objective-C specific.  If we change fpic.cmake, we would probably need to follow up with function-sections.cmake, hidden.cmake, etc.  This might be a case where it makes sense to add all of them at once.

The alternative would be maintaining a separate set of fortran flags in `polly/fflags/*.cmake`, with a new set of top level toolchains to explicitly specify fortran flags as needed, although that seems to add to the combinatorial explosion.

Please advise.


headupinclouds:
This change does not alter the toolchain-id, so the hunter cache for toolchains using the flag will be inconsistent.

PR #296, opened 2018-11-29
Support OS X Command Line tools without Xcode

bkotzz:
The current script sets the developer root by doing `xcode-select --print-path`. I have one machine with Xcode installed, and one with just the command line tools installed, no Xcode. 

On the machine with Xcode:
```
$ xcode-select --print-path
/Applications/Xcode.app/Contents/Developer
```

And if I ask for the SDK path:
```
$ xcrun --sdk macosx --show-sdk-path
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk
```

And on the machine with just command line tools:
```
$ xcode-select --print-path
/Library/Developer/CommandLineTools
```

And the SDK path:
```
$ xcrun --sdk macosx --show-sdk-path
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk
```

To support these setups, this change adds a case for when the root is in the command line tools directory.

Note: We *may* be able to just use `xcrun`, without having to add hardcoded strings to the end of `xcode-select --print-path`. But I'm not entirely sure

PTAL @ruslo

### DIFF
--- a/os/osx.cmake
+++ b/os/osx.cmake
@@ -52,9 +52,22 @@ if(_is_empty)
   endif()
 endif()
 
+# With a full Xcode install, typically `xcode-select -print-path` is something like:
+#     /Applications/Xcode.app/Contents/Developer
+#
+# But with just Xcode command line tools installed, the path is:
+#     /Library/Developer/CommandLineTools
+#
+# In the CommandLineTools case, the SDKs folder is at a different relative path.
+if(XCODE_DEVELOPER_ROOT MATCHES ".*CommandLineTools.*")
+    set(SDK_RELATIVE_PATH "SDKs")
+else()
+    set(SDK_RELATIVE_PATH "Platforms/MacOSX.platform/Developer/SDKs")
+endif()
+
 set(
     CMAKE_OSX_SYSROOT
-    "${XCODE_DEVELOPER_ROOT}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${OSX_SDK_VERSION}.sdk"
+    "${XCODE_DEVELOPER_ROOT}/${SDK_RELATIVE_PATH}/MacOSX${OSX_SDK_VERSION}.sdk"
     CACHE STRING "System root for OSX" FORCE
 )
 
--- a/os/osx.cmake
+++ b/os/osx.cmake
@@ -52,12 +52,6 @@ if(_is_empty)
   endif()
 endif()
 
-if(XCODE_DEVELOPER_ROOT MATCHES ".*CommandLineTools.*")
-    set(SDK_RELATIVE_PATH "SDKs")
-else()
-    set(SDK_RELATIVE_PATH "Platforms/MacOSX.platform/Developer/SDKs")
-endif()
-
 set(
     CMAKE_OSX_SYSROOT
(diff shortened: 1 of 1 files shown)


os/osx.cmake
```diff
@@ -52,9 +52,22 @@ if(_is_empty)
...
 endif()
 
+# With a full Xcode install, typically `xcode-select -print-path` is something like:
+#     /Applications/Xcode.app/Contents/Developer
+#
+# But with just Xcode command line tools installed, the path is:
+#     /Library/Developer/CommandLineTools
+#
+# In the CommandLineTools case, the SDKs folder is at a different relative path.
+if(XCODE_DEVELOPER_ROOT MATCHES ".*CommandLineTools.*")
```
ruslo: I think this condition is not very robust. What about checking if `"${XCODE_DEVELOPER_ROOT}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${OSX_SDK_VERSION}.sdk"` path exists and if it's not exist try `"${XCODE_DEVELOPER_ROOT}/SDKs/MacOSX${OSX_SDK_VERSION}.sdk"`?

ruslo: > Works for me, thanks.

Does it mean you will update this pull request?

bkotsopoulossc: Yes, will do

os/osx.cmake
```diff
@@ -58,6 +58,21 @@ set(
...
+if(NOT EXISTS "${CMAKE_OSX_SYSROOT}")
+  set(
+      CMAKE_OSX_SYSROOT
+      "${XCODE_DEVELOPER_ROOT}/SDKs/MacOSX${OSX_SDK_VERSION}.sdk"
+      CACHE STRING "System root for OSX" FORCE
+  )
+endif()
+
 if(NOT EXISTS "${CMAKE_OSX_SYSROOT}")
   polly_fatal_error("${CMAKE_OSX_SYSROOT} not exists")
```
ruslo: Now this error will not be accurate. If SDK is not installed it will emit error message "/Library/Developer/CommandLineTools/SDKs/MacOSXa.b.sdk not exists" however in most cases it should be "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSXa.b.sdk not exists".

I think it's better to do:
```cmake
set(__osx_sysroot_suggestion_1 "${XCODE_DEVELOPER_ROOT}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${OSX_SDK_VERSION}.sdk")
set(__osx_sysroot_suggestion_2 "${XCODE_DEVELOPER_ROOT}/SDKs/MacOSX${OSX_SDK_VERSION}.sdk")

if(EXISTS "${__osx_sysroot_suggestion_1}")
  set(__osx_sysroot "${__osx_sysroot_suggestion_1}")
elseif(EXISTS "${__osx_sysroot_suggestion_2}")
  set(__osx_sysroot "${__osx_sysroot_suggestion_2}")
else()
  # error message with both __osx_sysroot_suggestion_1 and __osx_sysroot_suggestion_2
endif()

# set CMAKE_OSX_SYSROOT to __osx_sysroot
```

bkotzz: Thanks for the suggestion, done.

PR #299, opened 2018-12-04
Add OS X xcrun fallback for finding SDK path

bkotzz:
Relates to and builds off of https://github.com/ruslo/polly/pull/296.

I have a Mac with the following results:
```
$ xcode-select -p
/Library/Developer/CommandLineTools
```

```
$ xcrun --show-sdk-version
10.14.1
```

```
$ ls -al /Library/Developer/CommandLineTools/SDKs
total 8
drwxr-xr-x  4 root  wheel  136 Nov 28 16:08 .
drwxr-xr-x  5 root  admin  170 Nov 28 16:08 ..
drwxr-xr-x  7 root  wheel  238 Oct 18 05:52 MacOSX.sdk
lrwxr-xr-x  1 root  wheel   10 Nov 28 16:08 MacOSX10.14.sdk -> MacOSX.sdk
```

```
$ xcrun --show-sdk-path
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

So, if the toolchain doesn't set the OS X version specifically, the current polly script gets it from xcrun, and uses this to build the path. But the SDK for the version reported by xcrun doesn't exist in the SDK folder. But, xcrun can give the full path to the SDK, and this path is valid and does exist.

So, this change adds another fallback if a valid SDK can't be found, which uses the absolute path from xcrun.

PTAL @ruslo


os/osx.cmake
```diff
@@ -69,12 +69,24 @@ set(
     "${XCODE_DEVELOPER_ROOT}/SDKs/MacOSX${OSX_SDK_VERSION}.sdk"
 )
 
+# If OSX_SDK_VERSION is not set, the SDK version is computed using
+# xcrun. However, it is possible for the version reported by xcrun to not
+# be present in the SDKs folder. In that case, xcrun can also give the
+# full path to a valid SDK.
+execute_process(
```
ruslo: We only need to run it if `__osx_sysroot_suggestion_1` and `__osx_sysroot_suggestion_2` not present, it make sense to move it to `else()` block:
```cmake
if(EXISTS ...)
elseif(EXISTS ...)
else()
  execute_process(...)
endif()
```

os/osx.cmake
```diff
@@ -69,12 +69,24 @@ set(
...
+    COMMAND xcrun --show-sdk-path
+    OUTPUT_VARIABLE __osx_sysroot_suggestion_3
```
ruslo: Please check exit code, it should be `0`. See `RESULT_VARIABLE` usage above.

ruslo:
Thanks, applied with small tweak: https://github.com/ruslo/polly/commit/35e86d6bbf6df8aefe66daa76a7d4a1eecf383c4

PR #300, opened 2018-12-12
Fix weird iOS (simulation toolchain) apps still getting signed with ios-nocodesign-*.

phaazon:
Colleagues of mine noticed that even with the `ios-nocodesign-11-2-dep-9-3` (simulation), the apps still got signed.

Please tell me whether this is the right place to do this change. I’ve used [this stack overflow response](https://stackoverflow.com/a/39901677) as a source.

\o/


ruslo:
Reverted https://github.com/ruslo/polly/commit/6f5ebf010d1e2432636a4f3c9c5742ec748cc8a8

It's breaking iOS framework build, example:
* https://travis-ci.org/ruslo/polly/builds/477364473#L409-L427

phaazon:
@ruslo I created an issue for that and I showed how I fixed it.

PR #308, opened 2019-02-01
Clang-7 Explicit Toolchain

arrtchiu:
Couple of things need review:
1. Naming convention: Switches to libc++ but name doesn't indicate that
2. Path to header is specific to where I've got those headers, I'm not sure how to more elegantly find them

### DIFF
--- a/clang-7-nocxx.cmake
+++ b/clang-7-nocxx.cmake
@@ -1,0 +1,25 @@
+# Copyright (c) 2016-2017, Ruslan Baratov
+# Copyright (c) 2017, David Hirvonen
+# All rights reserved.
+
+if(DEFINED POLLY_CLANG_7_CMAKE_)
+  return()
+else()
+  set(POLLY_CLANG_7_CMAKE_ 1)
+endif()
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
+
+polly_init(
+    "clang 7"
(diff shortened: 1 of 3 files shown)


clang-7-nocxx.cmake
```diff
@@ -0,0 +1,22 @@
...
+polly_init(
+    "clang 7"
+    "Unix Makefiles"
+)
+
+include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_common.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/library/std/libcxx.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/compiler/clang-7.cmake")
+
+include_directories(SYSTEM /usr/lib/llvm-7/include/c++/v1)
```
ruslo: How you're installing Clang? I've downloaded Clang 7 from this page:
- http://releases.llvm.org/download.html#7.0.1

and it works fine without extra hardcoded paths. I'm using `clang-libcxx` toolchain.

ruslo: Just for the info, `clang-libcxx98` added: https://github.com/ruslo/polly/commit/f12c23dba4f59378a63564f66b3d03963ae82636

arrtchiu:
Thanks to your help and some symlinks I don't think this toolchain makes sense anymore - closing :)

PR #309, opened 2019-02-07
Add macOS toolchains for 10.12 Deployment Target

remyjette:


bin/detail/toolchain_table.py
```diff
@@ -614,6 +614,9 @@ def verify(self):
       Toolchain('osx-10-13-dep-10-10', 'Xcode', osx_version='10.13'),
       Toolchain('osx-10-13-dep-10-10-cxx14', 'Xcode', osx_version='10.13'),
       Toolchain('osx-10-13-dep-10-10-cxx17', 'Xcode', osx_version='10.13'),
+      Toolchain('osx-10-13-dep-10-12', 'Xcode', osx_version='10.13'),
```
ruslo: It's missing

bin/detail/toolchain_table.py
```diff
@@ -622,6 +625,9 @@ def verify(self):
       Toolchain('osx-10-14-dep-10-10', 'Xcode', osx_version='10.14'),
       Toolchain('osx-10-14-dep-10-10-cxx14', 'Xcode', osx_version='10.14'),
       Toolchain('osx-10-14-dep-10-10-cxx17', 'Xcode', osx_version='10.14'),
+      Toolchain('osx-10-14-dep-10-12', 'Xcode', osx_version='10.14'),
```
ruslo: Missing too

bin/detail/toolchain_table.py
remyjette: Whoops, thanks!

bin/detail/toolchain_table.py
remyjette: fixed, thanks!

PR #313, opened 2019-03-06
Add support for Visual Studio 16 2019

gegles:


bin/detail/toolchain_table.py
```diff
@@ -375,6 +375,12 @@ def verify(self):
           arch='amd64',
           vs_version='15'
       ),
+      Toolchain(
+          'vs-16-2019',
+          'Visual Studio 16 2019',
+          arch='amd64',
```
ruslo: It should be `arch=x86` or generator should be `Visual Studio 16 2019 Win64`.

gegles: As per the [cmake 3.14 release notes](https://cmake.org/cmake/help/v3.14/release/3.14.html#id3), the new "Visual Studio 16 2019" is a bit different as in, it does not support the " Win64" and other variant suffix in the generator's name but instead forces you to use the -A flag....

gegles: This was my very quick attempt to get something working.... Maybe you could take a quick look into how best to support the -A flag?

ruslo: > As per the cmake 3.14 release notes, the new "Visual Studio 16 2019" is a bit different as in, it does not support the " Win64" and other variant suffix in the generator's name but instead forces you to use the -A flag....

Okay, I see. I'm downloading Visual Studio 2019, will experiment.

> This was my very quick attempt to get something working

`arch` was not used for Visual Studio generators in fact, it's mostly for the consistency.

gegles: Perfect, thanks for looking into it!

ruslo: If you change `arch` to `arch='x86'` it will be a correct configuration and I can merge it now.

gegles: Ok, you got it! 

Not sure it completely makes sense anymore (since now it will probably default to the systems arch, so most likely 64bit).

ruslo: > Not sure it completely makes sense anymore (since now it will probably default to the systems arch, so most likely 64bit).

`CMAKE_GENERATOR_PLATFORM` did the trick. `vs-16-2019` should build `x86` and `vs-16-2019-win64` should build `amd64`:
- https://github.com/ruslo/polly/commit/8ad34681924113ed701eeb26b55c28358e0adaa6

PR #314, opened 2019-03-25
VS /permissive- toolchains

dvirtz:


flags/vs-nonpermissive.cmake
```diff
@@ -0,0 +1,12 @@
...
+
+if(DEFINED POLLY_FLAGS_VS_NONPERMISSIVE_CMAKE_)
+  return()
+else()
+  set(POLLY_FLAGS_VS_NONPERMISSIVE_CMAKE_ 1)
+endif()
+
+include(polly_add_cache_flag)
+
+polly_add_cache_flag(CMAKE_CXX_FLAGS_INIT "/permissive-")
```
ruslo: If there is no C++ macro to check that this flag is enabled, then we should modify `HUNTER_TOOLCHAIN_UNDETECTABLE_ID` variable, e.g.:
- https://github.com/ruslo/polly/blob/c7462593865acbc0557fae6d461607a274dc6e36/flags/vs-z7.cmake#L17

ruslo: > If there is no C++ macro ...

If such macro exists then it should be added to Hunter:
- https://github.com/ruslo/hunter/blob/v0.23.142/scripts/predefined.list

dvirtz: done

dvirtz:
Thanks

flags/vs-nonpermissive.cmake
```diff
@@ -0,0 +1,16 @@
...
+
+# There is no macro to detect this flags on toolchain calculation so we must
+# mark this toolchain explicitly.
+list(APPEND HUNTER_TOOLCHAIN_UNDETECTABLE_ID "/permissive-")
```
ruslo: Tiny follow-up: https://github.com/ruslo/polly/commit/d28d406412f445a6e34d867e19720b4f729c8904

PR #316, opened 2019-04-03
Add C++14 and C++17 toolchains for Visual Studio 2019

remyjette:


bin/detail/toolchain_table.py
```diff
@@ -406,12 +406,36 @@ def verify(self):
...
+          arch='x86',
+          vs_version='16'
+      ),
       Toolchain(
           'vs-16-2019-win64',
           'Visual Studio 16 2019',
           arch='amd64',
           vs_version='16'
       ),
+      Toolchain(
```
GTValentine: do you want to add ninja support as well?

remyjette: I know nothing about ninja, so not in this PR.

vs-16-2019-win64.cmake
```diff
@@ -10,7 +10,7 @@ endif()
 include("${CMAKE_CURRENT_LIST_DIR}/utilities/polly_init.cmake")
 
 polly_init(
-    "Visual Studio 16 2019"
+    "Visual Studio 16 2019 Win64"
```
ruslo: Missed it. Thanks for the fix.

PR #320, opened 2019-04-22
add android-ndk-r17-api-24-arm64-v8a-clang-pic-libcxx14

headupinclouds:


ruslo:
Do we need it? I think PIC applied to Android toolchains by default. I have tested `android-ndk-r17-api-24-arm64-v8a-clang-libcxx14`:
```
/.../android-ndk-r17c/toolchains/llvm/prebuilt/linux-x86_64/bin/clang++ \
    ... -fPIC ... \
    -o CMakeFiles/foo.dir/foo.cpp.o -c /.../foo.cpp
```

I guess it's because the only way to use C++ code is to create shared `*.so` library and load it by Java code.

headupinclouds:
I guess it is redundant then.  I'll close it.

PR #328, opened 2019-07-03
update Emscripten environment variable and cmake path

Boris-Rasin:


compiler/emscripten.cmake
```diff
@@ -9,13 +9,12 @@ endif()
 
 include(polly_fatal_error)
 
-string(COMPARE EQUAL "$ENV{EMSCRIPTEN}" "" _is_empty)
+string(COMPARE EQUAL "$ENV{EMSDK}" "" _is_empty)
```
ruslo: Do we need `EMSCRIPTEN` for old SDK and `EMSDK` for the new SDK? In this case for compatibility, it's better to check both. I'm not using Emscripten actively so I would really appreciate if you leave some links to documentation and comments.

Boris-Rasin: $EMSDK was added on Oct 26, 2016 [here](https://github.com/emscripten-core/emsdk/commit/52dd54e4b65b78ca1aba9bf79e1c8c18055581ec). Not sure what the deal with $EMSCRIPTEN is, it seems to be defined for some versions but not the others [here](https://github.com/emscripten-core/emsdk/blob/master/emsdk_manifest.json).

ruslo: > $EMSDK was added on Oct 26, 2016 here

Toolchain was added before:
- https://github.com/ruslo/polly/commits/master/compiler/emscripten.cmake

> In this case for compatibility, it's better to check both

Seems like a right approach, it should be not so hard to test `EMSDK` and then `EMSCRIPTEN`.

ruslo:

PR #329, opened 2019-09-11
Add iOS 12.4 support

john-forrest:
Update of the iOS12.0 support we use to target iOS12.4 for Xcode10.3
usage.


john-forrest:
Added to wrong repo

PR #341, opened 2020-02-13
added lots of ios-13-2-dep-9-3 toolchains (individual devices, signed and unsigned)

belveder79:
Hi, added new toolchain without bitcode. Thx, clemens


belveder79:
and more toolchains...

belveder79:
Awesome, thx. Just a quick question: did you ever encounter that running a
build for all devices, say armv7, armv7s and arm64, fails, while individual
ones work out perfectly?

I ran the “device” toolchain, for which I get a compiler error. But running
toolchains individually work out. It is not a memory issue, and I have no
idea...

Ruslan Baratov <notifications@github.com> schrieb am So. 16. Feb. 2020 um
21:50:

ruslo:
> did you ever encounter that running a
build for all devices, say armv7, armv7s and arm64, fails, while individual
ones work out perfectly?

Yes, it happens sometimes. Not all CMake code is suitable for a multi architecture build, not all C++ code as well. E.g. `try_compile` will run only for a single architecture, and if the results of this operation will be architecture-dependent, you will have errors while compiling the untested architecture. Same problem if you save the results generated by `configure_file` C++ source.

Also, note that for a single architecture build we will set `CMAKE_SYSTEM_PROCESSOR`, for the multi architecture we can't:
- https://github.com/ruslo/polly/blob/99c7b708fed6a17488e11ca087a2beabf42e4dad/os/iphone.cmake#L157-L165

Examples:
- https://github.com/hunter-packages/curl/commit/72243ea2dc71629422d9d693e5b8ee8f5afd2a12
- https://github.com/hunter-packages/curl/commit/4ff2384b47719e2e8de4dc82cdf5d28e36149952
- https://github.com/hunter-packages/opencv/issues/28

PR #349, opened 2020-08-28
Add ios-dep-11 and ios-dep-12 toolchains

remyjette:
Also removed check for IOS_SDK_VERSION VERSION_LESS 10.0 in the
ios-dep-10 toolchain


ruslo (approved):
Looks good, thank you 👍

PR #350, opened 2020-10-09
ios 14 toolchains added

belveder79:
as usual, toolchains for IOS


ruslo (approved):
Looks good, thank you 👍

PR #351, opened 2020-12-04
Fix macOS 11.0 toolchains

remyjette:
Apple ended the 10.x series and went to 11.0 so these toolchain files
are incorrect.

### DIFF
--- a/bin/detail/toolchain_table.py
+++ b/bin/detail/toolchain_table.py
@@ -819,9 +819,9 @@ if platform.system() == 'Darwin':
       Toolchain('osx-10-15-dep-10-10-cxx14', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-10-cxx17', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-12-cxx17', 'Xcode', osx_version='10.15'),
-      Toolchain('osx-11-0', 'Xcode', osx_version='10.16'),
-      Toolchain('osx-11-0-cxx17', 'Xcode', osx_version='10.16'),
-      Toolchain('osx-11-0-dep-10-10-cxx17', 'Xcode', osx_version='10.16'),
+      Toolchain('osx-11-0', 'Xcode', osx_version='11.0'),
+      Toolchain('osx-11-0-cxx17', 'Xcode', osx_version='11.0'),
+      Toolchain('osx-11-0-dep-10-10-cxx17', 'Xcode', osx_version='11.0'),
       Toolchain('linux-gcc-x64', 'Unix Makefiles'),
   ]
 
--- a/bin/detail/toolchain_table.py
+++ b/bin/detail/toolchain_table.py
@@ -819,9 +819,9 @@ if platform.system() == 'Darwin':
       Toolchain('osx-10-15-dep-10-10-cxx14', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-10-cxx17', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-12-cxx17', 'Xcode', osx_version='10.15'),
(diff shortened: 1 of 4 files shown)


bin/detail/toolchain_table.py
```diff
@@ -819,9 +819,9 @@ def verify(self):
       Toolchain('osx-10-15-dep-10-10-cxx14', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-10-cxx17', 'Xcode', osx_version='10.15'),
       Toolchain('osx-10-15-dep-10-12-cxx17', 'Xcode', osx_version='10.15'),
-      Toolchain('osx-10-16', 'Xcode', osx_version='10.16'),
```
ruslo, line 822: There is no need to remove old toolchains. Just add new ones.

ruslo (changes requested):

bin/detail/toolchain_table.py
remyjette: Even though the 10.16 SDK doesn't exist? These toolchains won't ever do anything since Apple ended the 10.x series with 10.15 as the last entry

ruslo: I haven't touched macOS for a while, but I doubt that all the Xcode instances with all machines with SDK 10.16 installed suddenly stopped working in one moment. Just add new toolchains on top of the others. I really don't see any issue here.

ruslo: E.g. here:

- https://docs.travis-ci.com/user/reference/osx/#macos-version

I can even see `osx_image: xcode7.3 | Xcode 7.3.1 | 7D1014 | macOS 10.11 | 1.8.0_92-b14`

remyjette: My point is 10.16 SDK never existed in the first place. Apple has never released a 10.16 SDK, so it's not possible for anyone to be using this toolchain. There are no "Xcode instances with SDK 10.16 installed." And given that that Apple has ended 10.x, there never will be a 10.16 SDK.

Just doing some cleanup because this toolchain has no reason to exist as Apple decided to name this version of the SDK "11.0" rather than "10.16"

ruslo: > My point is 10.16 SDK never existed in the first place. Apple has never released a 10.16 SDK, so it's not possible for anyone to be using this toolchain. There are no "Xcode instances with SDK 10.16 installed."

`osx-10-16-*` was added here: https://github.com/ruslo/polly/pull/346

They were never working?

remyjette: Sadly no.

I created that PR in anticipation of (what I thought would be) the 10.16 release - at the time, my project wouldn't accept changes to our open source dependencies until *after* they had been merged upstream, and I was trying to get ahead of it so the toolchain would already be in place when Apple released the new SDK.

But after nearly 20 years of 10.x, Apple decided to switch to 11.0 and make me look very silly. So this PR is correcting my mistake.

ruslo: > > They were never working?

> Sadly no

Please do not add non-working toolchains. As I said, I was testing myself the macOS stuff before, but I'm not doing it right now. Having broken configuration in a set is a pretty bad idea.

bin/detail/toolchain_table.py
```diff
@@ -819,9 +819,9 @@ def verify(self):
...
-      Toolchain('osx-10-16-cxx17', 'Xcode', osx_version='10.16'),
-      Toolchain('osx-10-16-dep-10-10-cxx17', 'Xcode', osx_version='10.16'),
+      Toolchain('osx-11-0', 'Xcode', osx_version='10.16'),
```
ruslo, line 822: > osx_version='10.16'

Should it be `11.0`?

ruslo (changes requested):

ruslo (approved):
Looks good, thank you 👍

